PR #21, opened 2015-12-28
pep8 clean up

vnitinv:
My first pull request to pyangbind with [pep8](https://www.python.org/dev/peps/pep-0008/) cleanup using tool [autopep8](https://github.com/hhatto/autopep8)

### DIFF
--- a/lib/base.py
+++ b/lib/base.py
@@ -20,72 +20,73 @@ limitations under the License.
 
 import copy
 
+
 class PybindBase(object):
 
-  __slots__ = ()
+    __slots__ = ()
 
-  def elements(self):
-    return self._pyangbind_elements
+    def elements(self):
+        return self._pyangbind_elements
 
-  def __str__(self):
-    return str(self.elements())
+    def __str__(self):
+        return str(self.elements())
 
-  def get(self, filter=False):
-    def error():
-      return NameError, "element does not exist"
-    d = {}
-    # for each YANG element within this container.
-    for element_name in self._pyangbind_elements:
-      element = getattr(self, element_name, error)
-      if hasattr(element, "yang_name"):
-        # retrieve the YANG name method
-        yang_name = getattr(element, "yang_name", error)
-        element_id = yang_name()
-      else:
-        element_id = element_name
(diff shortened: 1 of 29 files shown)


robshakir:
Hi,

It's pretty difficult to tell what this pull request actually does, other than changing the indentation style from two spaces to four. Can you give me some view on the motivation for this change, and a bit of a run through of what the other changes that the tool made are?

Cheers,
r.

vnitinv:
If we run "pep8 ." under folder pyangbind, we can notice many type of pep8 violation 

for ex:
```
./tests/xpath/01-list_leaflist/run.py:72:21: E712 comparison to True should be 'if cond is True:' or 'if cond:'
./tests/xpath/01-list_leaflist/run.py:72:80: E501 line too long (84 > 79 characters)
./tests/xpath/01-list_leaflist/run.py:73:7: E127 continuation line over-indented for visual indent
./tests/xpath/01-list_leaflist/run.py:75:3: E111 indentation is not a multiple of four
./tests/xpath/01-list_leaflist/run.py:78:1: E302 expected 2 blank lines, found 1
./tests/xpath/01-list_leaflist/run.py:78:17: E231 missing whitespace after ','
```
I do understand 90% of changes relate to space (2 to 4). I thought to clean it up as python developer uses IDE like PyCharm, IDLE etc which follows pep8 rule of 4 space. So in future once we have many contributor toward pyangbind there wont be mix of 2 & 4 spaces.

We can ignore this pull request as of now.

vnitinv:
Rob, I am closing this pull request inline with #22 but I still go by my point "python developer uses IDE like PyCharm, IDLE etc which follows pep8 rule of 4 space. So in future once we have many contributor toward pyangbind there wont be mix of 2 & 4 spaces."

robshakir:
@vnitinv 

I agree that this is not-PEP8 complaint, and there are people that might use IDEs/development environments where indentation style cannot be set. I'd expect users of pyangbind not to edit the generated code manually - so I'm not clear that a *user* would care. If there are others that want to contribute to pyangbind that find this a problem, then we can reconsider.

PR #22, opened 2015-12-31
Formatting fixes to support PEP8

robshakir:
@vnitinv 

I propose that we merge this pull-req to support general PEP8 compliance. Commit comment below shows what is not complied with - hopefully this is something that you can set in your editor to ignore (Sublime Text 3, with SublimeLint and the PEP8 module has this ability). Comments welcome.

	* As per Nitin's suggestion -- code is now mostly compliant with
	  the guidelines in PEP8. Exceptions:
	  	*	Indentation is 2 spaces, not 4 (ignore: E111, E114)
	  	* 'Natural' indentation is used for continuation (Ignore: E127,
	  		 E128)

### DIFF
--- a/lib/base.py
+++ b/lib/base.py
@@ -20,6 +20,7 @@ limitations under the License.
 
 import copy
 
+
 class PybindBase(object):
 
   __slots__ = ()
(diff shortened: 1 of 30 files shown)


robshakir:
To be clear, I propose we merge this *instead of* #21.

PR #23, opened 2016-01-01
Very minor spelling mistake (Seralization -> Serialization)

mpainenz:


robshakir:
Thanks for this -- and thanks for maintaining the British English spelling :-) -- I'll merge this now.

mpainenz:
No worries, we who live in the colonies respect the queen's english.

I was contemplating adding an example of deserialization as well to your example, that is, once I work out how it's done. When I first read through the description, I wasn't sure if deserialization was included yet. I'm pleased to see that serialise.py includes it.

robshakir:
Hi Mark,

:+1: on the Queen's English! :-)

I added a very brief example into the README for the deserialisation case -- however, it seemed like there might be some ambiguity there still -- so I added a worked example - which is a link to https://gist.github.com/robshakir/c4221228b209e0abe847.

Hopefully this helps!

Cheers,
r.

mpainenz:
Hi Rob,

Thanks for this. Just to give you some more feedback, the difficulties I experienced so far using this library were as follows.

* How to best create a new project that includes the required pyangbind files which are located in the lib directory (base.py, yangtypes.py etc). I ended up taking a copy of these files and adding them to my own project. I'm new to python, so it's likely that I've missed a better way to configure python. Perhaps I needed a pythonpaths entry for pyangbind, I'm not sure.

* How to create and add entries to list objects which have been defined in the class bindings file which pyangbind creates.

For the second item above, I found that your tests area was very handy to work this out. And your new example also clearly explains this very well.

I'll let you know if anything else comes up, but thanks for adding that example, that's very helpful.

Cheers,
Mark

robshakir:
Hi Mark,

Thanks for this — it’s useful to understand what’s not intuitive, such that I can work on fixing that going forward :-)

On 6 January, 2016 at 2:43:41 PM, Mark Paine (notifications@github.com) wrote:

How to best create a new project that includes the required pyangbind files which are located in the lib directory (base.py, yangtypes.py etc). I ended up taking a copy of these files and adding them to my own project. I'm new to python, so it's likely that I've missed a better way to configure python. Perhaps I needed a pythonpaths entry for pyangbind, I'm not sure

The way that I’ve ended up doing this at the moment is:

import sys
sys.path.append('/path/to/pyangbind/')
import lib.yangtypes as yangtypes
However, this is clearly not optimal — as per Nitin’s suggestion, I have a branch (https://github.com/robshakir/pyangbind/tree/packaging) that packages the project such that it can be imported into PyPI. I did not push it to the live PyPI server just yet - but you can install it using:

 pip install -i https://testpypi.python.org/pypi pyangbind
If you’re happy to use this from the test server (for thee moment, it’ll be uploaded once Nitin looked at the branch), then you can rather just do:

from Pyangbind import yangtypes
I need to check some of the packaging - as looking at the test package - it seems like there might be some file duplication in there.

On 6 January, 2016 at 2:43:41 PM, Mark Paine (notifications@github.com) wrote:

How to create and add entries to list objects which have been defined in the class bindings file which pyangbind creates.

This was something that I’ve struggled a little with - as such, what you want to do is to have the equivalent to:

d = {}
d['foo'] = some_new_obj
but because d[‘foo’] is a particular class instance in this case, we need some way to be able to initialise the entry, such that it can be manipulated. To me, it seemed like having .add(key) and .delete(key) were probably the best names - because it’s not really the same as .append() on a normal list. A Python dict doesn’t have a analogous method that one could steal the name from.

Hopefully the packaging will make the import more logical in the future :-)

Cheers, r.

mpainenz:
Hi Rob, Thanks for the information. 

I found that if I changed the location of the required files (yangtypes), that I got errors during json serialization, this was because the map of classes references the lib module:

`if map_val in ["lib.yangtypes.RestrictedClass"]:`

If I were to install pyangbind from PyPi, I wonder if that would still work, or would I get a similar error perhaps because the class name may now be `pyangbind.lib.yangtypes.RestrictedClass`?

I also wonder if going forward, it might be better to have the import statements in the class binding files which pyandbind generates to use the `from pyangbind` import, rather than the `from lib` import.

I hope this makes sense, what I think is the aim is to get the class output to work without needing to edit it, by just installing pyangbind from PyPi.

In regards to your comments about the need to instantiate the list items, that makes sense. It took me a while to realize that the Class output structure was not 'normal' Python lists etc. So once I had worked out what was going on, it was a lot clearer. Again, I put that down to learning a new language.

Thanks,
Mark

robshakir:
Hi Mark,

Sorry - not sure whether you saw - but I fixed this as I resolved #24 - it should be a lot more robust than it would have been previously.

I've migrated all our code here @ Jive to using this approach too - so the test coverage should be pretty OK!

Cheers,
r.

mpainenz:
Thanks Rob,

I'll give it a try soon.

PR #39, opened 2016-02-29
Issue when creating a list item, when a named set is used, and only one key value is supplied which doesn't contain a space

mpainenz:
Hi Rob,

Minor bug error when creating a keyed list item with a named key (key=key value) where a single key value is used, and no space character is contained in the Key Value supplied

The following worked: 

```
pybind_obj = pybindobj().somelist.add("somevalue")
pybind_obj = pybindobj().somelist.add(keyfield="some value")
pybind_obj = pybindobj().somelist.add(keyfield1="somevalue", keyfield2="somevalue")
```

The following didn't work (no space, single key, named key):

`pybind_obj = pybindobj().somelist.add(keyfield="somevalue")`

Cheers,
Mark

### DIFF
--- a/pyangbind/lib/yangtypes.py
+++ b/pyangbind/lib/yangtypes.py
@@ -571,7 +571,7 @@ def YANGListType(*args, **kwargs):
               raise KeyError("YANGList key must contain all key elements (%s)"
                                 % (self._keyval.split(" ")))
 
-          elif " " in self._keyval and named_set:
+          elif named_set:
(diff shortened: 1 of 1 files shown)


robshakir:
Thanks Mark - I think this as I was only thinking about the case where the key value contained a space but the key dictionary is built the same way in both cases. I'll merge this, then add a unit test.

Thanks,
r.

mpainenz:
Thanks Rob :)

Issue #41, opened 2016-03-08
Identityref In pyangbind


ashishk1994:
Hi ,

I have following yang file.
```
 identity address-family {
    description
      "Base identity from which address
       families are derived.";
  }
 identity lcaf {
    base address-family;
    description
      "address family.";
  }
  identity source-dest {
    base lcaf;
    description
      "Source/Dest LCAF type.";
 }
 typedef address-family-ref {
    type identityref {
      base address-family;
    }
    description
      "address family reference.";
  }
  grouping address {
    description
      "Generic address.";
    leaf address-type {
      type address-family-ref;
      mandatory true;
      description
        "Type of the address.";
    }
  }
```
Here address -> address-type is of type address-family-ref which is identityref deriving the address-family. Now I want to set  leaf address-type in address as source-dest type  but It is giving me value error. source-dest is not directly inheriting from the address-type, Is that the case?
Error:
```
Traceback (most recent call last):
  File "generate.py", line 7, in <module>
    Object.eid.address_type = 'source-dest-key-lcaf'
  File "/home/ashish/odl_test/bindings.py", line 70, in _set_address_type
    raise ValueError("""address_type must be of a type compatible with base=RestrictedClassType(base_type=unicode, restriction_type="dict_key", restriction_arg={u'foo:lcaf': {}, u'lcaf': {}},), is_leaf=True, yang_name="address-type", parent=self, path_helper=self._path_helper, extmethods=self._extmethods, register_paths=True""")
ValueError: address_type must be of a type compatible with base=RestrictedClassType(base_type=unicode, restriction_type="dict_key", restriction_arg={u'foo:lcaf': {}, u'lcaf': {}},), is_leaf=True, yang_name="address-type", parent=self, path_helper=self._path_helper, extmethods=self._extmethods, register_paths=True
```

ashishk1994:
I have located the issue in the pybind.py file. Actually It is not recursively adding the derived classes in base classes. For example if we have hierarchy as below 
A -> B -> C -> D
then currently we are restricting it as A = [B] , B = [C], C = [D] but It should contain all the classes like A = [B, C , D] , B = [C, D], C = [D], D = [].

ashishk1994:
please review : https://github.com/robshakir/pyangbind/pull/42

PR #42, opened 2016-03-09
Recursive addition of classes in building identities.

ashishk1994:
Fixes Bug: https://github.com/robshakir/pyangbind/issues/41


robshakir:
Ashish,

Thanks for the work here -- I think this needs a little more examination. There are two problems that I think I can identity with the solution here:

1. The `for` loop that you've added appears to break the `while` loop that makes sure that we repeat the resolution process if the base does not exist.
2. I don't think that you can be sure that all the values have been added to the identity at the time that you do the inheritance. It seems like it needs to be done after all identities have been processed.

I'll take a look, as I do think #41 is an issue.

Cheers,
r.

robshakir:
Hi Ashish,

I just resolved this and merged the fix, thanks for the pull-req. The recursive code I wrote resolves the two issues above.

Cheers,
r.

ashishk1994:
@robshakir  Thanks a lot for adding my example :)

robshakir:
No problem - I just cut v0.3.2, so this includes this fix.

PR #48, opened 2016-03-20
Fix when a YANG Restriction specifies a range including 0, IE... '0..…

mpainenz:
…max'. 0 was being evaluated as an invalid value, when it is inside the range.

### DIFF
--- a/pyangbind/lib/yangtypes.py
+++ b/pyangbind/lib/yangtypes.py
@@ -831,7 +831,7 @@ def YANGDynClass(*args, **kwargs):
           break
         except:
           pass  # don't worry, move on, plenty more fish (types) in the sea...
-      if not type_test:
+      if not type_test and type(type_test).__name__ != 'RestrictedClass':
         # we're left alone at midnight -- no types fit the arguments
         raise TypeError("did not find a valid type using the argument as a" +
                             "hint")


mpainenz:
Hi Rob, 

Here is an example of some code that doesn't work under the current master branch:

```
            value = 0

            t = YANGDynClass(value,
                             base=[
                                 #RestrictedClassType(base_type=unicode,restriction_type="dict_key", restriction_arg={u'unlimited': {}},),
                                 RestrictedClassType(base_type=np.int16, restriction_dict={'range': [u'0..max']}),],
                             default=unicode("unlimited"),
                             is_leaf=True,
                             yang_name="max-ingress-calls",
                             parent=self,
                             #path_helper=self._path_helper,
                             #extmethods=self._extmethods,
                             register_paths=True,
                             namespace='http://genband.com/sbc/endpoints',
                             defining_module='endpoints',
                             yang_type='union')
```

robshakir:
Hi Mark,

ACK on the bug, but not on the fix. The fix you've added basically breaks any `union` that has a class that has a restriction on it (e.g., `int16` with `range` will now just not be checked).

This was actually a more subtle bug `if not val` in Python evaluates to `True` if the value is 0 - so any `if` on the value was saying that this was not a valid value, even though it might have been. Thanks for the bug report - I've fixed it in https://github.com/robshakir/pyangbind/commit/6e7a14c61f22634a0d3775885bbfdcc3777ab4c3 - please give this a go.

Cheers,
r.

robshakir:
Closing pull-req, please re-open if this is still an issue.

mpainenz:
Thanks Rob, I had tried 

`if type_test == False:`

However that still triggered with the int value of zero. I don't know how `if type_test is False:` works, but I'm glad you could find a better way than mine. I suspected you would be able to improve upon it, as it was a rather cumbersome way to avoid the issue (not to mention the even bigger bug I introduced).

Thanks for that,

Mark

robshakir:
Essentially in Python there can only ever be one instance of `True` and `False` (as per `None`). The test `is False` then checks whether the return value is the same `False` rather than testing it against something else that could evaluate to False (e.g., 0).

No problems,
r.

Issue #52, opened 2016-03-28
Expose information about field constraints


timmartin:
I'm writing some code that's attempting to make automated changes to a Pyangbind object tree (for model-based testing). In order for my code to make valid changes, it needs to know the types of fields, their constraints and which of them are non-configurable status fields. Obviously I can extract this information from the Yang model separately, but it seems a shame that I've already got the Pyangbind object in front of me but I can't query it to find metadata.

One obvious problem is how to expose this. The ideal would be something like:

```
>>> config.users.username.is_config
True
>>> config.users.date_created.is_config
False
```
But if this ends up applied to non-leafs it could get confused with a real Yang member. Perhaps it actually belongs on the type (since it's a property of the type and not the value stored in it), so it would be something like:

```
>>> type(config.users.username).is_config
True
```

robshakir:
Hi Tim,

`config` isn't currently saved, but doing this would be quite easy - one needs to add to the argument string of each generated dynamic type (through changing `pybind.py` -  particularly the argument string that is built), and then subsequently adding this info in `YANGDynType` as a `kwarg`. 

Restrictions for a leaf are already accessible through `_restriction_dict` of any leaf that has a restriction specified on it. This does not include built-in restrictions (e.g., `uint32` range restrictions) where they are implemented by the base type that PyangBind exploits.

The type of the field is stored in the `_yang_type` of the generated class.

Let me know if the `_restriction_dict`/`_yang_type` info is what you're looking for -- I can assist with how you'd implement the `config` one if that's not clear.

Cheers,
r.

timmartin:
I haven't had time to look into the `_restriction_dict` and `_yang_type`, but if the information is there I'm sure I can adapt it to my needs. I've had a crack at implementing the `_is_config` field.

PR #53, opened 2016-03-31
Nodes now report whether they are configurable

timmartin:
I've added support for nodes to report whether they are configurable, with a field called `_is_config`. See #52.


robshakir:
Looking at this now :microscope: - the CI checks failing are based on another issue that I resolved in `master` (change to OpenConfig models).

robshakir:
LGTM :+1:, merging - I'll add another couple of checks in the unit test once merged. 

Thanks for this!

PR #56, opened 2016-04-05
Fixed a typo in _key_to_native_key_type

timmartin:


robshakir:
@timmartin 

Thanks for this! I added this for consuming applications so there's no consumption directly inside PyangBind - hence it wasn't caught by any tests. Thanks for the catch.

I'll merge this request now.

r.

PR #59, opened 2016-04-25
While experimenting with pyangbind and current openconfig yang models,

cyrusd:
I found that it was necessary to modify the get_element method (specifically to
process release/models/mpls/openconfig-mpls.yang).

### DIFF
--- a/pyangbind/plugin/pybind.py
+++ b/pyangbind/plugin/pybind.py
@@ -1311,8 +1311,11 @@ def get_element(ctx, fd, element, module, parent, path,
   # dynamically generates a class.
 
   # Find element's namespace
-  namespace = element.i_orig_module.search_one("namespace").arg \
-                    if hasattr(element, "i_orig_module") else None
+  namespace = None
+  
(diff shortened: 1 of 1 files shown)


robshakir:
Hi,

Thanks for the PR! I validated that this was an issue -- looking at the fix, whilst this one works, I'm not sure that it's quite right. In the case that `i_orig_module` does not have a `namespace` attribute, then we are in a submodule - to this end, the check that is needed is to validate that this is a submodule, and then find the namespace of the main module. Not doing this means that we get the namespace set to `None`, which is fine for general operation, but if one is serialising to XML, it's potentially incorrect. This also highlights an issue where the `defining_module` is set incorrectly (to the submodule name rather than the module name) which is incorrect for IETF JSON serialisation.

To this end, I'll commit a fix with the above implemented - the PR is much appreciated though, even though I won't merge it!

Thanks,
r.

cyrusd:
Excellent, thanks!  I admit to being a bit of a YANG novice, so definitely
appreciate you digging into this a bit and coming up with a proper fix :)

robshakir:
No problems at all, thanks for reporting it! :-)

r.

PR #61, opened 2016-04-27
Issue with Min values for Integer types with a Restriction/Range specified

mpainenz:
Hi Rob,

As you can see from the commit comments:

> Corrected values for the ranges of signed Integer types to include the correct minimum value
> Added test case for minimum value of an integer with a restriction/range
> Added test case for 'min' alias value which is currently broken

I've included an example of a still broken case where if you specify a restriction that uses the 'min' keyword, it does not work. It's currently commented out, as it breaks the test.

```
leaf restricted-ueight-min-alias {
    type int8 {
        range "min..max";
    }
}
```

### DIFF
--- a/pyangbind/plugin/pybind.py
+++ b/pyangbind/plugin/pybind.py
@@ -138,24 +138,24 @@ class_map = {
         "pytype": YANGBool
     },
     'int8': {
-        "native_type": "RestrictedClassType(base_type=int, restriction_dict={'range': ['-127..127']}, int_size=8)",
+        "native_type": "RestrictedClassType(base_type=int, restriction_dict={'range': ['-128..127']}, int_size=8)",
         "base_type": True,
-        "pytype": RestrictedClassType(base_type=int, restriction_dict={'range': ['-127..127']}, int_size=8)
+        "pytype": RestrictedClassType(base_type=int, restriction_dict={'range': ['-128..127']}, int_size=8)
     },
     'int16': {
-        "native_type": "RestrictedClassType(base_type=int, restriction_dict={'range': ['-32767..32767']}, int_size=16)",
+        "native_type": "RestrictedClassType(base_type=int, restriction_dict={'range': ['-32768..32767']}, int_size=16)",
         "base_type": True,
-        "pytype": RestrictedClassType(base_type=int, restriction_dict={'range': ['-32767..32767']}, int_size=16)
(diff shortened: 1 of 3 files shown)


mpainenz:
By the way, that Travis CI is really neat, what a great tool

robshakir:
Hey Mark,

Good catch, thanks! In this case, I'd managed to just set them to the same value for positive and negative, in addition to your tests, I added some bounds tests so that this doesn't break again :-) Thanks very much for the work here.

The `min` keyword was just not specified in the regexps that match the ranges. These are used to check whether a range is a single value, or actually a specified set of ranges. I added it there, I also removed the check that didn't allow `min..max` to be specified. I think I'd misinterpreted 6020 for this one - hence it's now gone.

In terms of where these values are translated into the numerical values - essentially they're not - a specification of `1..max` becomes `(1, None)` and the range check simply skips the maximum value test if the value of the range maximum is set to `None`. This happens in [this `range_check` function](https://github.com/robshakir/pyangbind/blob/master/pyangbind/lib/yangtypes.py#L217-L237).

I've merge the pull request to a branch, added to it, and I'll merge that branch now - I just wanted to be able to fix the `min` before committing this. `master` will have this done in a few moments.

I agree on Travis -- it's great! Minimal time to get it running too :-D

Thanks again for the contribution!

r.

mpainenz:
Ah right, I was looking in plugin.py... I'm slowly getting the whole picture

Really glad to be able to contribute. This is a really interesting project Rob, I'm learning a lot digging into it.

Thanks again,
Mark

robshakir:
I'll write some documentation around how the code is laid out in Pyangbind when I get a moment -- I've not done this thus far based on the fact that it's mostly been me! Thanks again for the contributions.

Essentially, `pybind.py` does all the logic to try and get the YANG tree that Pyang produces to being mapped into some class with all the restrictions that are required encoded as arguments. This always calls `YANGDynType` from `lib.yangtypes` which is a meta-class that wraps some other class.

All classes that aren't natively supported by something in Python (see the `class_map` in `pybind.py`) are then defined in `lib.yangtypes`. Some of these wrap other types, e.g., `RestrictedClassType` is a wrapper around the input type itself. So generally, if it's an issue with how the YANG is mapped, it's in `pybind.py` if it's an issue with how the type works, it's probably in `lib.yangtypes`.

If it's an issue with how it gets dumped, or loaded, then `lib.serialise`, and XPATH things are in `lib.xpathhelper`. :-)

Thanks again,
r.

mpainenz:
Sounds great, thanks Rob.

PR #65, opened 2016-05-04
YANG "bits" type

cyrusd:
Hi Rob,

I took a swing at some code to implement support for the YANG `bits` type...  I have tested it and looked at serialized XML output to confirm that the representation matches the description in RFC-6020.  

I also am including a little shell script I threw together to generate handy Python class files from a directory of YANG models...  Seems to do the trick with the set of openconfig models I have at the moment.  Let me know if this is useful :)


robshakir:
Hi Cyrus,

Thanks for the contribution.

In your code, if I use the bit naming method, then the value of the class never gets updated. For example:

```python
>>> 
>>> from bindings import bits
>>> b = bits()
>>> 
>>> b.container.b1
bitarray('000000')
>>> b.container.b1[0] = True
>>> b.container.b1
bitarray('100000')
>>> b.container.b1.flag1 = False
>>> b.container.b1
bitarray('100000')
```

This is because when one runs `b1.flag1` then you are simply setting the `flag1` attribute on the `b1` instance of `bitarray.bitarray`. This isn't actually related to the contents of the bitarray.

Additionally, where you have the `__setitem__` code for the named bits, then `__setitem__` is not actually called, rather `__setattr__` is. If you need to overload `__setattr__` there are a number of considerations to look at in `YANGDynClass` to ensure that `_set()` gets called correctly for this leaf.

One other issue is how this type will work when one includes it in a `union` type - with the other types, the `__init__` for the type will always take an argument (named `v` in the generated bindings) that indicates the initial value of the class - when this returns `ValueError` then the next type in the union is tried. With your code currently, I think that we will end up throwing away any `args` that are given to `YANGBits` and rather initialise as all 0s. This ends up being an issue for unions, since we'll always consider the supplied value valid.

Also. I think there's more to do with respect to serialisation - I started to look at this (we need to add tests in `tests/serialise/ietf-json-serialise`, `tests/serialise/json-serialise`, `tests/serialise/ietf-json-deserialise` and `tests/serialise/json-deserialise` for data of the 'bits' type). It also looks like there'll be work to do to support:

```yang 
typedef derived-bits {
  type bits;
}

typedef bitsinunion {
  type union {
    type bits;
    type string;
  }
}
```

Since these types will not return `orig_yangt` as `bits` (a lovely side effect of YANG!).

Apologies, I don't mean this review to be disparaging, but I want to make sure that we add types that work as users would expect to the code!

Thanks,
r.

robshakir:
Thanks for the `y2c` script - it looks good! I wonder whether we could merge this separately, and ideally add support for `--split-class-dir` into it.

Thanks again,
r.

cyrusd:
Hey Rob - thanks for the feedback!  I'll retract this pull request and send you another with just the y2c script.  I'll keep poking at the `bits` type; a number of the IETF models use it.  Thanks for pointing out some of the deficiencies so far - I'll include better tests next time as well!

PR #77, opened 2016-07-16
Added property to reserved_name

dbarrosop:


dbarrosop:
Looks like the CI error has nothing to do with my patch:

    AssertionError: Could not get file 404 from GitHub (response: https://raw.githubusercontent.com/openconfig/public/master/release/models/bgp/openconfig-bgp-multiprotocol.yang)

robshakir:
Hi David!

Thanks for this -- indeed the CI error is due to a change in the OpenConfig repository. I fixed that, and will merge this now.

Much appreciated,
r.

PR #79, opened 2016-07-17
Added support for YANG Notifications and the related test unit.

lgifre:
Hi Rob,

I added support for YANG notifications to PyangBind and the related test unit.
Please, take a look and let me know if there is something missing/incorrect.

Best,
Lluis

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ tests/pyvirtualenv
 dev/*
 tmp/*
 *.PEP8-ERRORS
+/.project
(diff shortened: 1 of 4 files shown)


robshakir:
Hi Lluis,

Great, this looks fantastic, thanks. I'll merge the pull-req now. Apologies, the CI failed due to some changes in external modules.

One thing that we should add to the tests is serialisation and deserialisation of a Notification.

Cheers,
r.

PR #84, opened 2016-08-04
corrected import for built in function 'property' to remove ambiguity…

krishn2014:
Corrected import for built in function 'property' to remove ambiguity if a yang tree has a node named 'property'.


robshakir:
I've merged this to `master`. Thanks for the contribution!

r.

PR #100, opened 2016-11-05
Fixed UTs that use unittest.main not supporting the -k argument

### LINKED ISSUE #99: Unit tests using unittest module fail when DEL=true set in yang_tests.sh
If you set `DEL=true` in `yang_tests.sh`, then the following tests are reported as failing:

- tests/misc
- tests/presence
- submodules

This is because these tests use `unittest.main`, which rejects the `-k` argument used to indicate that the output should not be deleted.

timmartin:
This fixes #99. I'm pulling out the `-k` argument and passing everything else on to `unittest`, which is the simplest fix. Perhaps it would be better to pass no args by default and add specific arguments that are supported, in order to make the command-line options more consistent between the different `run.py` scripts, though I don't think this is a big deal right now.

### DIFF
--- a/tests/integration/openconfig-interfaces/run.py
+++ b/tests/integration/openconfig-interfaces/run.py
@@ -125,7 +125,10 @@ class PyangbindOCIntf(unittest.TestCase):
 
 if __name__ == '__main__':
   setup_test()
-  T = unittest.main(exit=False)
+  args = sys.argv
+  if '-k' in args:
+    args.remove('-k')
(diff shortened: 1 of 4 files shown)


robshakir:
Hi Tim,

Thanks for spotting this and the fix -- I built on your code a bit to ensure that -k actually did what it was intended to do in these cases. I'll close this, but do re-open it if you don't think the commits I'll merge do what you expect.

Thanks,
r.

PR #103, opened 2016-11-19
Resolved simple syntax errors in Python3

timmartin:
This change has fixes for the basic syntax errors that occur with Python 3. I'm doing some work to make the code Python 3 compatible and this is the first step. It's not enough on its own, but merging these simple changes will mean there's less noise when applying later more involved patches.


robshakir:
Hi Tim,

Apologies, completely missed this. I'll pull this in after I've submitted a few outstanding changes I have locally.

Thanks for the contribution!

r.

Issue #109, opened 2017-02-03
openconfig-bgp.yang serializing to json fails 


sanabby:
Generation of python classes for openconfig-bgp yang model works fine.

Initializing the BGP objects from generated classes and serializing it to json fails. This is because of the following traceback which shows AttributeError: Unmapped type.

Traceback (most recent call last):
  File "load.py", line 12, in <module>
    l=pyb.encode(ocbgp)
  File "/Library/Python/2.7/site-packages/pyangbind/lib/serialise.py", line 529, in encode
    self._preprocess_element(obj, mode="ietf"))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
  File "/Library/Python/2.7/site-packages/pyangbind/lib/serialise.py", line 532, in default
    return pybindJSONEncoder().default(obj, mode="ietf")
  File "/Library/Python/2.7/site-packages/pyangbind/lib/serialise.py", line 137, in default
    val = self.map_pyangbind_type(pyc, orig_yangt, obj, mode)
  File "/Library/Python/2.7/site-packages/pyangbind/lib/serialise.py", line 197, in map_pyangbind_type
    return self._preprocess_element(obj.get(), mode=mode)
  File "/Library/Python/2.7/site-packages/pyangbind/lib/serialise.py", line 57, in _preprocess_element
    nd[k] = self._preprocess_element(d[k], mode=mode)
  File "/Library/Python/2.7/site-packages/pyangbind/lib/serialise.py", line 57, in _preprocess_element
    nd[k] = self._preprocess_element(d[k], mode=mode)
  File "/Library/Python/2.7/site-packages/pyangbind/lib/serialise.py", line 57, in _preprocess_element
    nd[k] = self._preprocess_element(d[k], mode=mode)
  File "/Library/Python/2.7/site-packages/pyangbind/lib/serialise.py", line 57, in _preprocess_element
    nd[k] = self._preprocess_element(d[k], mode=mode)
  File "/Library/Python/2.7/site-packages/pyangbind/lib/serialise.py", line 57, in _preprocess_element
    nd[k] = self._preprocess_element(d[k], mode=mode)
[...]

PR #110, opened 2017-02-03
support for Decimal object added. This fixes serializing Decimal objects as unicode objects in json

sanabby:
This PR fixes the issue, openconfig-bgp.yang serializing to json fails #109


robshakir:
Thanks for the PR. I'll merge an alternate solution.

This one has the issue that it does not check whether there's a specific mapping type that pyangbind is carrying before checking the native type. This is needed for when there is wrapping around a native class.

Please watch for a commit in the next few minutes with this, once I've added the testing.

PR #116, opened 2017-03-07
Fixing bug with empty type.

GrandVizierOlaf:
load_ietf_json was not properly handling the empty type. You can tell that it was never called because set_method was being called, but it hadn't been defined yet.

This change checks for [None] instead of None.

I had to update the existing test to include the now-included {'empty': True} part of the expected dictionary. I included some debugging with jsondiff for when the JSON doesn't match, but could remove that if you'd prefer.

### DIFF
--- a/pyangbind/lib/serialise.py
+++ b/pyangbind/lib/serialise.py
@@ -462,7 +462,8 @@ class pybindJSONDecoder(object):
         if chk._is_keyval is True:
           pass
         elif chk._yang_type == "empty":
-          if d[key] == None:
+          if d[key] == [None]:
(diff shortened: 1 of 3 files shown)


robshakir:
Hi!

Thanks for the PR - it looks great. Could you please sign the Google CLA mentioned by the bot above? I can then go ahead and merge this change.

Thanks again for your contribution.

Thanks,
r.

robshakir:
Unfortunately, I can't merge this contribution until such time as the CLA has been signed. We'll have to look to solve these issues elsewhere without that.

robshakir:
Thanks for this contribution -- unfortunately, as the CLA wasn't signed, it couldn't be merged -- a fix has been applied through another PR.

r.

PR #118, opened 2017-03-19
Using OrderedDict by default as some devices need ordering

dbarrosop:
I am using pyangbind to translate OpenConfig into native configuration and I am running into issues as some devices require ordering when entering IP addresses. For example, the following configuration will fail:

```
interface Ethernet2.1
    encapsulation dot1q vlan 1
    ip address 172.20.0.1/24 secondary
    ip address 192.168.1.1/24
napalm_base.exceptions.MergeConfigException: Error [1000]: CLI command 17 of 40 'ip address 172.20.0.1/24 secondary' failed: could not run command [Primary address must be assigned first]
```

The solution is as simple as using always `OrderedDict` instead of when only `user_ordered = True`.

### DIFF
--- a/pyangbind/lib/serialise.py
+++ b/pyangbind/lib/serialise.py
@@ -54,7 +54,7 @@ class pybindJSONEncoder(json.JSONEncoder):
         for k in d:
             if isinstance(d[k], dict) or isinstance(d[k], OrderedDict):
                 nd[k] = self._preprocess_element(d[k], mode=mode)
-                if isinstance(d, OrderedDict):
+                if getattr(d, "_user_ordered", False):
                     nd[k]['__yang_order'] = index
             else:
                 nd[k] = self.default(d[k], mode=mode)
--- a/pyangbind/lib/yangtypes.py
+++ b/pyangbind/lib/yangtypes.py
@@ -468,7 +468,7 @@ def TypedListType(*args, **kwargs):
 def YANGListType(*args, **kwargs):
   """
     Return a type representing a YANG list, with a contained class.
-    A dict or ordered dict is used to store the list, and the
+    An ordered dict is used to store the list, and the
     returned object behaves akin to a dictionary.
 
     Additional checks are performed to ensure that the keys of the
@@ -503,10 +503,10 @@ def YANGListType(*args, **kwargs):
 
     def __init__(self, *args, **kwargs):
       self._ordered = True if user_ordered else False
-      if user_ordered:
-        self._members = collections.OrderedDict()
-      else:
-        self._members = dict()
+      self._members = collections.OrderedDict()
+
+      self._members._user_ordered = True if user_ordered else False
+
       self._keyval = keyname
       if not type(listclass) == type(int):
         raise ValueError("contained class of a YANGList must be a class")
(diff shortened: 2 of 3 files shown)


dbarrosop:
BTW, I know the tests are failing. I will look into them but I wanted to first send the PR to see if this is something that you will want to merge or not.

robshakir:
Hi David,

Interesting one - I'm not sure I quite understand the issue here. Is the problem here in the ordering of which address is primary in the dict? I had two ways to address this previously:

- Don't make the primary address deterministic.
- Use an extension to indicate which the primary address is which is then used by the config generation template/code.

It'd be good to discuss wider, because there's some implications AFAICS of changing this from `dict`->`OrderedDict`, especially where folks are using `isinstance()`

Cheers!
r.

dbarrosop:
> Don't make the primary address deterministic.

That probably has two consequences:
1. You will get annoying and unnecessary diffs as the secondary keywords just shifts around.
2. It might also have an impact on some protocols as the primary IP might be used to form adjacencies or to select which network to advertise.

> Use an extension to indicate which the primary address is which is then used by the config generation template/code.

Not sure I understand, I wrote an augmentation to set that field and that's working fine. The problem is when a iterate over the list I don't have any way of ensuring I process the IP addresses in the same order I entered them unless I use an OrderedDict.

> It'd be good to discuss wider, because there's some implications AFAICS of changing this from dict->OrderedDict, especially where folks are using isinstance()

One would hope people is using `is_container|leaf` or `_yang_type` to do type checks but I understand, that's why I wanted to discuss first before looking into CI.

The thing is that I wrote an open source framework that can parse any native configuration/show_command and transform it into a YANG model and vice versa so treating this as an individual case (and potentially others) might be problematic. It would be much better if we could fix it in the source.

sjtarik:
To David's point, JUNOS RPC calls also expect the fields in parameters are in the order given by the Junos RPC Yang file.

robshakir:
@sjtarik - I think this is a different issue. This implies that we need to store somewhere what order we had the statements in from the file and then serialise this with that order. This is unsafe (since JSON objects' keys don't maintain their order).

The issue that @dbarrosop raises is actually related to lists AIUI, essentially, we'd like to always use an ordered dictionary such that the order of insertion matters.

@dbarrosop - let's try and go ahead and merge this, I think it seems OK, and we have had this issue elsewhere.

dbarrosop:
>  I think this is a different issue

yes, that's correct. Those are two different issues. The issue where fields have to be ordered (oh, the joy), I haven't yet figured out up to what extent it's really a problem. The only case I have found so far is that the `name` of the interface has to be the very first parameter when sending the object to the device but I am sure there will be others and, probably, on other vendors too.

> let's try and go ahead and merge this, I think it seems OK, and we have had this issue elsewhere.

Thanks! I will fix CI.

dbarrosop:
Ok, I think this is it. Quick notes, as you were using OrderedDict sometimes to know if the order was relevant, in order to replicate the old behavior I added a `_user_ordered` attribute to the OrderedDict object and used that instead. So, now, instead of `isinstance(d, OrderedDict)` we can do `getattr(d, "_user_ordered", False)` and achieve the same result. No need to change `isinstance(d, dict)` as `OrderedDict` evaluates `True`.

Oh, I also added `values` and `items` as those were missing.

robshakir:
awesome, thanks @dbarrosop! LGTM, and thanks for signing the Google CLA a while back. Will merge this now.

PR #121, opened 2017-04-10
Add original_module in addition to defining_module

dbarrosop:
When a YANG module uses a grouping from another module, `defining_module` is set as the module of the "importer". For those cases, `original_module` will keep the information of the module where the imported grouping was defined for reference.

For example, the grouping `bgp-top`, when used by `network_instances/network_instance[key]/protocols/protocol[key]` will have:

* `defining_module='openconfig-network-instance'`
* `original_module='openconfig-bgp'`

### DIFF
--- a/pyangbind/plugin/pybind.py
+++ b/pyangbind/plugin/pybind.py
@@ -945,6 +945,7 @@ def get_children(ctx, fd, i_children, module, parent, path=str(),
           class_str["arg"] += ", is_keyval=True"
         class_str["arg"] += ", namespace='%s'" % i["namespace"]
         class_str["arg"] += ", defining_module='%s'" % i["defining_module"]
+        class_str["arg"] += ", original_module='%s'" % i["original_module"]
         class_str["arg"] += ", yang_type='%s'" % i["origtype"]
         class_str["arg"] += ", is_config=%s" % (i["config"] and parent_cfg)
         classes[i["name"]] = class_str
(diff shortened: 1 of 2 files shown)


dbarrosop:
I am closing this PR as I noticed this is not enough for my use case, I might have to do something similar with the path.

pyangbind/plugin/pybind.py
```diff
@@ -1622,3 +1630,4 @@ def get_element(ctx, fd, element, module, parent, path,
 
     this_object.append(elemdict)
   return this_object
+
```
robshakir: Could you remove this blank line? Aiming not to make whitespace changes far from the original changes.

pyangbind/plugin/pybind.py
```diff
@@ -1611,7 +1618,8 @@ def get_element(ctx, fd, element, module, parent, path,
         "choice": choice,
         "register_paths": register_paths,
         "namespace": namespace,
-        "defining_module": defining_module
+        "defining_module": defining_module,
+        "original_module": element.i_orig_module.arg,
```
robshakir: Can `i_orig_module` ever be `None`, or can `element` not have `i_orig_module` as an attribute? 

The case above seems to show that it can be, but this one just references it directly.

pyangbind/plugin/pybind.py
```diff
@@ -1399,6 +1405,7 @@ def get_element(ctx, fd, element, module, parent, path,
           "register_paths": register_paths,
           "namespace": namespace,
           "defining_module": defining_module,
+          "original_module": element.i_orig_module.arg,
```
robshakir: Per other comment, can `i_orig_module` never be `None`?

robshakir:
@dbarrosop - no problems, sorry, review was in-flight when you changed.

dbarrosop:
> review was in-flight when you changed

No worries, will take them into account if/when I resubmit it.

Thanks! :)

PR #133, opened 2017-07-01
Add some PY3 improvements

ktbyers:
There are more things to do, but this moves closer to PY3 support.


ktbyers:
I am fixing the test failures.

ktbyers:
@robshakir These test failures happen in the current master branch (when I run it on my current machine)

The first error is:

```
/home/travis/build/robshakir/pyangbind/tests/serialise/ietf-json-serialise/ietf-json-serialise.yang:67: warning: the escape sequence "\." is unsafe in double quoted strings - pass the flag --lax-quote-checks to avoid this warning
/home/travis/build/robshakir/pyangbind/tests/serialise/ietf-json-serialise/ietf-json-serialise.yang:67: warning: the escape sequence "\." is unsafe in double quoted strings - pass the flag --lax-quote-checks to avoid this warning
```

Which I can make go away by single quoting the pattern in `ietf-json-serialise.yang`


The second set of errors are these:

```
/home/travis/build/robshakir/pyangbind/tests/integration/openconfig-interfaces/include/openconfig-vlan-types.yang:84: warning: the escape sequence "\." is unsafe in double quoted strings - pass the flag --lax-quote-checks to avoid this warning
/home/travis/build/robshakir/pyangbind/tests/integration/openconfig-interfaces/include/openconfig-vlan-types.yang:84: warning: the escape sequence "\." is unsafe in double quoted strings - pass the flag --lax-quote-checks to avoid this warning
/home/travis/build/robshakir/pyangbind/tests/integration/openconfig-interfaces/include/openconfig-vlan-types.yang:100: warning: the escape sequence "\." is unsafe in double quoted strings - pass the flag --lax-quote-checks to avoid this warning
/home/travis/build/robshakir/pyangbind/tests/integration/openconfig-interfaces/include/openconfig-vlan-types.yang:100: warning: the escape sequence "\." is unsafe in double quoted strings - pass the flag --lax-quote-checks to avoid this warning
...
# more of same warning
```

Which I can also make go away in my repo by single quoting the pattern in the specified yang files and re-running the tests. 

Note, these files are not are included in the repo (i.e. are dynamically built which is why I didn't submit a PR on these fixes).

robshakir:
Thanks! Let me take a look at the failing tests and this PR over the next day or so. Apologies, been tied up in another YANG-related project that we're hoping to publish soon.

Thanks very much indeed for the work!

ktbyers:
Sounds good. I temporarily just fixed to `pyang==1.7.1` to work around the test failure issue.

pyangbind/lib/pybindJSON.py
```diff
@@ -17,11 +17,19 @@
...
+from pyangbind.lib.serialise import pybindIETFJSONEncoder
 import json
 import copy
 
+# Kludge to temporarily fix unicode references
+try:
+  unicode
+except NameError:
+  unicode = str
+
```
robshakir: Can you explain why this change is required? If we have unicode literals above, shouldn't we just convert over everything to being unicode literals and remove the references to `unicode`?

pyangbind/plugin/pybind.py
```diff
@@ -20,6 +20,7 @@
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+#from __future__ import unicode_literals
```
robshakir: Please remove this stale code.

pyangbind/plugin/pybind.py
```diff
@@ -38,6 +39,17 @@
 from pyang import statements
 from pyang import util
 
+# Temporrary kludge for Python3 support
```
robshakir: s/Temporrary/Temporary/

pyangbind/plugin/pybind.py
```diff
@@ -38,6 +39,17 @@
...
+try:
+  long
+except NameError:
+  long = int
```
robshakir: I don't think we can workaround this one, but please consider using this pattern:

```python
if sys.version_info > (3,):
  long = int
```

This makes it more explicit what we're checking for here.

pyangbind/plugin/pybind.py
```diff
@@ -309,20 +321,30 @@ def build_pybind(ctx, modules, fd):
...
+  ctx.pybind_common_hdr += "from bitarray import bitarray\n"
+  ctx.pybind_common_hdr += "import six\n"
+
+  # Python3 support
+  ctx.pybind_common_hdr += """
+# PY3 support of some PY2 keywords (needs improved)
+if six.PY3:
+ import builtins as __builtin__
+ long = int
+ unicode = str
```
robshakir: Per the above comment on `unicode`, can we simply move to unicode literals everywhere?

pyangbind/plugin/pybind.py
ktbyers: Sounds good.

pyangbind/lib/pybindJSON.py
ktbyers: There are two issues here:

1. Unicode literal strings which are fixed with the `from __future__ import unicode_literals`.
2. Type casting of strings as unicode in the code. Python3 has no `unicode` keyword so we can't do `unicode(some_string)`. The solution for this in Six Library is:

```
six.text_type(some_string)
```

Which resolves to `str` in Python3 (which is unicode) and `unicode` in Python2.

There were some references in this code which weren't apparent to me how to fix this way, however.

For example,

```
    'string': {
        "native_type": "unicode",
        "base_type": True,
        "quote_arg": True,
        "pytype": unicode
    },
```

You have this `native_type` with a name unicode as a string. So I just did a simpler solution and added the name `unicode` into Python3.

Open to the other fix, however, as it is a better way to do it (i.e. to typecast using the Six library and have no direct unicode references in the code).

pyangbind/plugin/pybind.py
ktbyers: Will remove.

pyangbind/plugin/pybind.py
ktbyers: Okay, added comment above.

ktbyers:
There are a couple of other changes I still need to make.

ktbyers:
@robshakir Okay, I think it should be updated now to reflect the review comments you added earlier. I consistently used the six library for the Python3 check.

robshakir:
Thanks! This is looking good.

One thing I think we probably want to check is that the tests can run for both python2 and python3 on a system where it's available -- with these changes we should ensure that we don't get py2/py3 specific regressions.

I think the best way to do this is via Travis for the moment, since this is able to run with multiple python versions -- could you add:

```
python:
  - "2.7"
  - "3.4"
```

To `.travis.yml` please? If you're interested in 3.2 and 3.3 these can also be added there.

Thanks again for all the work on this!

r.

ktbyers:
@robshakir Yes, this doesn't fully add Python3 support. It is a step on the way (so the travis test will probably fail on PY3)

My main focus is trying to get the generated code to be Py3 compliant for napalm-yang work. I still have a ways more to go on that.

robshakir:
OK! So let's leave regression just for py2.7 right now then. Merging. Thanks very much for the work here!

Issue #142, opened 2017-09-23
TypedListType not handled correctly by load_json


lamehost:
Hello,

Running the script below id expect it returns True, but it does not.
```python
import json
import openconfig_bindings
from pyangbind.lib import pybindJSON
from pyangbind.lib.serialise import pybindJSONDecoder

input = {
  "network-instance": {
    "global": {
      "name": "global", 
      "protocols": {
        "protocol": {
          "bgp 64496": {
            "bgp": {
              "neighbors": {
                "neighbor": {
                  "192.0.2.5": {
                    "neighbor-address": "192.0.2.5", 
                    "config": {
                      "auth-password": "super-secret-key", 
                      "peer-as": 64497, 
                      "description": "Friendly Peer"
                    }, 
                    "afi-safis": {
                      "afi-safi": {
                        "IPV4_UNICAST": {
                          "ipv4-unicast": {
                            "prefix-limit": {
                              "config": {
                                "shutdown-threshold-pct": 75, 
                                "restart-timer": 30.0, 
                                "max-prefixes": 5000
                              }
                            }
                          }, 
                          "apply-policy": {
                            "config": {
                              "import-policy": [
                                "AS64496_IN"
                              ],
                              "export-policy": [
                                "AS64406_OUT"
                              ]
                            }
                          }, 
                          "config": {
                            "enabled": True
                          }, 
                          "afi-safi-name": "IPV4_UNICAST"
                        }
                      }
                    }
                  }
                }
              }
            }, 
            "identifier": "bgp", 
            "name": "64496"
          }
        }
      }
    }
  }
}

instance_obj = openconfig_bindings.network_instances.network_instances()
output = pybindJSONDecoder.load_json(input, None, None, obj=instance_obj)
output = json.loads(pybindJSON.dumps(output))
print input == output
```

Reason is _serialize_'s _load_json_ method TypedListType objects during parsing.
Here's the relevant piece of code:
```python
      elif pybind_attr in ["TypedListType"]:
        if not overwrite:
          list_obj = getattr(obj, "_get_%s" % safe_name(key))()
          for item in d[key]:
            if item not in list_obj:
              list_obj.append(item)
          list_copy = []
          for elem in list_obj:
            list_copy.append(elem)
          for e in list_copy:
            if e not in d[key]:
              list_obj.remove(e)
          set_via_stdmethod = False
```

_list_obj_ is of YANGBaseClass type while _d[key]_ is regular list.
[...]

PR #144, opened 2017-10-18
fixes https://github.com/robshakir/pyangbind/issues/142

lamehost:


lamehost:
I signed it!

lamehost:
As far as i can see we need to update python within travis environment.
Can you take a look?

robshakir:
Thanks for the contribution!

It looks like the test that is failing is:

```
TESTING /home/travis/build/robshakir/pyangbind/tests/serialise/json-deserialise...
Traceback (most recent call last):
  File "/home/travis/build/robshakir/pyangbind/tests/serialise/json-deserialise/run.py", line 151, in <module>
    main()
  File "/home/travis/build/robshakir/pyangbind/tests/serialise/json-deserialise/run.py", line 119, in main
    "Load of object with all items not correct"
AssertionError: Load of object with all items not correct
```

Let me take a look at why this test might be failing.

robshakir:
This does look to be a test failure that is due to this PR. If you merge master to fix the external tests and check it can be debugged more. Let me know if you need some assistance with this.

JoseIgnacioTamayo:
As the branch was deleted, deleting also the PR.

PR #145, opened 2017-11-14
Python3 dict keyview to list fix in pybind.

dgjustice:
Added cast to list on `known_types` and `unresolved_t` to be compatible with Python3 dictionary view functions.


dgjustice:
I signed it!

robshakir:
Thanks for the commit. This looks good to me. Merging the PR.

PR #147, opened 2017-12-01
IETF serialization option to add default values if they are explicitly set

asloboda-cisco:
Retrying pull request 146 with another branch to pass CLA.


asloboda-cisco:
@robshakir do you have any means of checking why @googlebot doesn't accept CLA?  Is there manual processing involved?  We have a group, CLA was submitted and I accepted invitation into group and the commit is made with the same mail.  There's nothing else we can do.

robshakir:
@asloboda-cisco  -- apologies, @googlebot doesn't seem to check corporate CLAs. I can see that Cisco has signed this, so we're good to go here.

pyangbind/lib/serialise.py
```diff
@@ -508,20 +509,25 @@ def generate_element(obj, parent_namespace=None, flt=False):
...
           del d[yname]
       else:
         if flt and element._changed():
           d[yname] = element
-        elif not flt:
-          d[yname] = element
+        else:
+          if not flt:
+            d[yname] = element
+          elif with_defaults and element._default == element:
```
robshakir: This will only include defaults when they are explicitly set AFAICS. It seems like the check that you want to do here is `elif with_defaults and not element._default == False` since this is then a leaf that is specified to have a default.

pyangbind/lib/serialise.py
```diff
@@ -476,7 +476,8 @@ def load_ietf_json(d, parent, yang_base, obj=None, path_helper=None,
 
 class pybindIETFJSONEncoder(pybindJSONEncoder):
   @staticmethod
-  def generate_element(obj, parent_namespace=None, flt=False):
+  def generate_element(obj, parent_namespace=None, flt=False,
```
robshakir: Can you add to the docstring here to explain what `with_defaults` does please? Especially given the comment below where I don't think it matches what I'd expect.

pyangbind/lib/serialise.py
asloboda-cisco: That's exactly what I want to do, I only want to include values that were explicitly set.  How should I rename it then?

robshakir: I would suggest having an argument that takes an enumerated (`Enum` or `IntEnum`) which lets one specify the type of defaults addition that one wants. There are a few default reporting modes (e.g., those discussed in https://tools.ietf.org/html/rfc6243#section-2). I'd add an enumerated type that is something like `IF_SET` that does what you'd like. In the future we can extend this with other behaviours if they are required.

asloboda-cisco:
@robshakir there's an added dependency for enum.  It's your call if you want to add it or we can just use simple class.

robshakir:
I'm OK with the added dependency -- I think there are other places whereby this will be useful for having explicit specification of a behaviour.

Can you please add `enum` to `requirements.txt`? The tests are failing because the tests cannot import `enum`.

Thanks for the work here!

robshakir:
Thanks -- apologies, I missed the change here. This looks good, thanks for the work!

PR #152, opened 2018-01-03
Fix #151 load_ietf_json properly yang empty types

### LINKED ISSUE #151: Error at load_ietf_json with yang empty types
There is an error at loading in a container with several data and one of them is of type empty. For example:
  ...
  grouping volume-info {
    description "whatever";
    leaf device-bus {
      description "Type of disk-bus on which this disk is exposed to guest";
      type enumeration {
        enum ide;
        enum scsi;
      }
    }
    leaf ephemeral {
      type empty;
    }
  }

And load a json like:
    "volume-info":
    {
      "device-bus": "ide"
      "ephemeral ": null
    }

It seems to set True to device-bus

alf-tierno:
This intend to solve the problem observed when loading json (load_ietf_json) into an empty yang type object from a null json type value

### DIFF
--- a/pyangbind/lib/serialise.py
+++ b/pyangbind/lib/serialise.py
@@ -468,6 +468,10 @@ class pybindJSONDecoder(object):
           pass
         elif chk._yang_type == "empty":
           if d[key] == None:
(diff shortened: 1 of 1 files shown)


alf-tierno:
To be commited again with right author email

PR #153, opened 2018-01-03
Fix #151 load_ietf_json properly yang empty types

### LINKED ISSUE #151: Error at load_ietf_json with yang empty types

alf-tierno:
This intend to solve the problem observed when loading json (load_ietf_json) into an empty yang type object from a null json type value

### DIFF
--- a/pyangbind/lib/serialise.py
+++ b/pyangbind/lib/serialise.py
@@ -466,19 +466,21 @@ class pybindJSONDecoder(object):
         chk = get_method()
         if chk._is_keyval is True:
           pass
-        elif chk._yang_type == "empty":
-          if d[key] == None:
+        else:
+          val = d[key]
+          if chk._yang_type == "empty":
+            # A 'none' value in the JSON means that an empty value is set, 
+            # since this is serialised as [null] in the input JSON.
+            if val is None:
+              val = True
+            else:
+              val = None  # Ignore an input value different than null/empty
+          if val is not None:
             set_method = getattr(obj, "_set_%s" % safe_name(ykey), None)
             if set_method is None:
               raise AttributeError("Invalid attribute specified in JSON - %s"
                                       % (ykey))
-            set_method(True)
-        else:
-          set_method = getattr(obj, "_set_%s" % safe_name(ykey), None)
-          if set_method is None:
-            raise AttributeError("Invalid attribute specified in JSON - %s"
-                                    % (ykey))
-          set_method(d[key])
+            set_method(d[key])
         pybindJSONDecoder.check_metadata_add(key, d, get_method())
     return obj
 
--- a/pyangbind/lib/serialise.py
+++ b/pyangbind/lib/serialise.py
@@ -468,6 +468,10 @@ class pybindJSONDecoder(object):
           pass
         elif chk._yang_type == "empty":
           if d[key] == None:
+            set_method = getattr(obj, "_set_%s" % safe_name(ykey), None)
+            if set_method is None:
+              raise AttributeError("Invalid attribute specified in JSON - %s"
+                                      % (ykey))
             set_method(True)
         else:
           set_method = getattr(obj, "_set_%s" % safe_name(ykey), None)
--- a/pyangbind/lib/serialise.py
+++ b/pyangbind/lib/serialise.py
@@ -471,16 +471,17 @@ class pybindJSONDecoder(object):
           if chk._yang_type == "empty":
             # A 'none' value in the JSON means that an empty value is set, 
             # since this is serialised as [null] in the input JSON.
-            if val is None:
+            if val == [None]:
               val = True
             else:
-              val = None  # Ignore an input value different than null/empty
+              raise ValueError("Invalid value for empty in input JSON " + \
+                      "key: %s, got: %s" % (ykey, val))
           if val is not None:
             set_method = getattr(obj, "_set_%s" % safe_name(ykey), None)
             if set_method is None:
               raise AttributeError("Invalid attribute specified in JSON - %s"
                                       % (ykey))
-            set_method(d[key])
+            set_method(val)
         pybindJSONDecoder.check_metadata_add(key, d, get_method())
     return obj
 
--- a/tests/serialise/ietf-json-deserialise/run.py
+++ b/tests/serialise/ietf-json-deserialise/run.py
@@ -106,7 +106,8 @@ def main():
             "remote-identityref": "remote:stilton",
             "int64": 1,
             "restricted-string": "aardvark",
-            "decimal": Decimal('16.32')
+            "decimal": Decimal('16.32'),
+            "empty": True,
           }
       },
       "l2": OrderedDict(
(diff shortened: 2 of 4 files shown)


robshakir:
Hi!

Thanks for signing the CLA first off -- much appreciated.

I had a suggestion as to the way that this will be implemented, right now we end up with some duplication within the code due to the fact that you do a different check for `empty` above. I would propose doing this like:

```python
else:
        set_method = getattr(obj, "_set_%s" % safe_name(ykey), None)
        if set_method is None:
                raise AttributeError(...)

        val = None
        if chk._yang_type == "empty":
                # A 'none' value in the JSON means that an empty value
                # is set, since this is serialised as [null] in the input
                # JSON.
                if d[key] is None:
                        val = True
        else:
                val = d[key]

        if val is not None:
                set_method(val)
```

This would mean that you can avoid the `elif` clause and handle this only in the `else`. PTAL and let me know what you think.

Cheers,
r.

robshakir:
Can you also add a test that would fail for this please?

alf-tierno:
Sure! I will add a new commit to avoid code repetition.

On the other hand, what should it be the behaviour of providing a JSON value different than "null" for an “empty” YANG type? Current code will just ignore it at all as if this key/values were not provided. I think it will be better either 1) raise and exception, or 2) set the YANG object with True. If you agree tell me your preferred option (1 or 2) and I will also incorporate it.

Regards

alf-tierno:
This error seems not to be connected with the code. Can you please examine it and relaunch again? Thanks.

alf-tierno:
I have rebased these two commits. The problem that they fixed still applies. Can you please revise it?

robshakir:
I re-ran the testing and the tests in tests/serialise/ietf-json-serialise still fail (see the result here: https://travis-ci.org/robshakir/pyangbind/builds/343337546#L526).

Looking at the change too, I think the change in the `ietf-json-serialise` tests that you've made isn't valid according to RFC7951. The `empty` type should be serialised to be `[null]` rather than simply `null`: https://tools.ietf.org/html/rfc7951#section-6.9 -- i.e., the original JSON output is correct. This is actually where the test above is failing.

It also looks like there isn't a test for deserialise - and the code here fails (the `set_method` is called on `d[key]` rather than the rewritten value).

Here's a patch that fixes your branch -- if you can apply it, then push, then I can merge this PR:

```
From c546e1ff611816494273ffc77518e37afab7a077 Mon Sep 17 00:00:00 2001
From: Rob Shakir <robjs@google.com>
Date: Mon, 19 Feb 2018 10:35:48 -0800
Subject: [PATCH 3/3] Fix empty bugfix.

---
 pyangbind/lib/serialise.py                        | 7 ++++---
 tests/serialise/ietf-json-deserialise/run.py      | 6 ++++--
 tests/serialise/ietf-json-serialise/json/obj.json | 4 ++--
 tests/serialise/ietf-json-serialise/run.py        | 9 ++++++++-
 4 files changed, 18 insertions(+), 8 deletions(-)

diff --git a/pyangbind/lib/serialise.py b/pyangbind/lib/serialise.py
index 6b528f5..a2cec9a 100644
--- a/pyangbind/lib/serialise.py
+++ b/pyangbind/lib/serialise.py
@@ -471,16 +471,17 @@ class pybindJSONDecoder(object):
           if chk._yang_type == "empty":
             # A 'none' value in the JSON means that an empty value is set, 
             # since this is serialised as [null] in the input JSON.
-            if val is None:
+            if val == [None]:
               val = True
             else:
-              val = None  # Ignore an input value different than null/empty
+              raise ValueError("Invalid value for empty in input JSON " + \
+                      "key: %s, got: %s" % (ykey, val))
           if val is not None:
             set_method = getattr(obj, "_set_%s" % safe_name(ykey), None)
             if set_method is None:
               raise AttributeError("Invalid attribute specified in JSON - %s"
                                       % (ykey))
-            set_method(d[key])
+            set_method(val)
         pybindJSONDecoder.check_metadata_add(key, d, get_method())
     return obj
 
diff --git a/tests/serialise/ietf-json-deserialise/run.py b/tests/serialise/ietf-json-deserialise/run.py
index 9ed1fad..7bbc7a6 100755
--- a/tests/serialise/ietf-json-deserialise/run.py
+++ b/tests/serialise/ietf-json-deserialise/run.py
@@ -106,7 +106,8 @@ def main():
             "remote-identityref": "remote:stilton",
             "int64": 1,
             "restricted-string": "aardvark",
-            "decimal": Decimal('16.32')
+            "decimal": Decimal('16.32'),
+            "empty": True,
           }
       },
       "l2": OrderedDict(
@@ -128,7 +129,8 @@ def main():
     }
   }
   assert nobj.get(filter=True) == expected_get, "Deserialisation of " + \
-    "complete object not as expected"
+          "complete object not as expected, got: %s" % \
+          dumps(nobj.get(filter=True), indent=4, mode="ietf")
 
   pth = os.path.join(this_dir, "json", "nonexistkey.json")
   for i in [True, False]:
diff --git a/tests/serialise/ietf-json-serialise/json/obj.json b/tests/serialise/ietf-json-serialise/json/obj.json
index 1e75fb4..a7aa39c 100644
--- a/tests/serialise/ietf-json-serialise/json/obj.json
+++ b/tests/serialise/ietf-json-serialise/json/obj.json
@@ -53,7 +53,7 @@
                     "chicken"
                 ], 
                 "uint32": 1, 
-                "empty": null, 
+                "empty": [null], 
                 "int32": 1, 
                 "int16": 1, 
                 "string": "bear", 
@@ -79,4 +79,4 @@
     "ietf-json-serialise:augtarget": {
         "augment:augleaf": "teststring"
     }
-}
\ No newline at end of file
+}
diff --git a/tests/serialise/ietf-json-serialise/run.py b/tests/serialise/ietf-json-serialise/run.py
index a60096c..67f48bf 100755
--- a/tests/serialise/ietf-json-serialise/run.py
+++ b/tests/serialise/ietf-json-serialise/run.py
@@ -4,6 +4,7 @@ import os
 import sys
 import getopt
 import json
+import difflib
 from pyangbind.lib.serialise import pybindJSONEncoder, pybindIETFJSONEncoder
 from pyangbind.lib.pybindJSON import dumps
 from decimal import Decimal
@@ -101,7 +102,8 @@ def main():
   external_json = json.load(
                       open(os.path.join(this_dir, "json", "obj.json"), 'r'))
 
-  assert pybind_json == external_json, "JSON did not match the expected output"
+  assert pybind_json == external_json, "JSON did not match the expected output, " + \
+	"diff: %s\n" % diff(pybind_json, external_json)
 
   # Check that we can serialise a single element on its own.
   pybind_json = json.loads(json.dumps(
@@ -113,5 +115,10 @@ def main():
     os.system("/bin/rm %s/bindings.py" % this_dir)
     os.system("/bin/rm %s/bindings.pyc" % this_dir)
 
+def diff(a, b):
+    ja = json.dumps(a, indent=4, sort_keys=True).split("\n")
+    jb = json.dumps(b, indent=4, sort_keys=True).split("\n")
+    return '\n'.join(difflib.unified_diff(ja, jb))
+
 if __name__ == '__main__':
   main()
-- 
2.16.0.rc1.238.g530d649a79-goog
```

robshakir:
Thank you for making these changes. I'll merge this PR now.

r.

PR #138, opened 2017-08-20
More py3

dbarrosop:
This brings full py3 support and fixes a couple of bugs I found when running tests py3. I am still doing more tests with my own tooling (napalm-yang) to make sure this doesn't break anything that it's not in the tests but so far it looks good.

The bugfixes here are:

1. bitarrays don't support unicode
2. ordered lists without order specified had a strange logic that worked by pure luck. I replaced that logic to use an ordered dict when leading the json and using that order to create the ordered keys.
3. slots weren't used properly

PR #158, opened 2018-02-17
use ordereddict so we can process order of items consistently

dbarrosop:
As discussed here:

https://github.com/robshakir/pyangbind/pull/138/files#r144704978

the idea is to be able to process the elements in a consistent order mostly so you can generate consistently native configuration. Main use case for that is testing, it is very hard to test code that generates, for instance, native IOS configuration from a YANG model if the order keeps changing between tests.


robshakir:
Thanks for the PR -- I merged back master into this branch, and it LGTM. Bringing this in now.

PR #159, opened 2018-02-17
Initialize objects only when needed to lower memory footprint and cpu usage

dbarrosop:
YANG models are usually quite large and the tendency is to make them larger as the community tries to join them together to describe a full device configuration with a single data structure. This is a problem when using classes generated by pyangbind as each instance of an object has to be fully initialized even in the case where you only care about a tiny part of it.

The idea is to initialize the attributes only when needed. This brings crazy improvements. For instance, https://github.com/napalm-automation/napalm-yang tests under this branch takes considerably less time to run:

```
regular pyangbind

tests for ios-interfaces
    - 4.88 seconds

All tests
    - 89.65 seconds 

optimized branch

tests for ios-interfaces
    - 3.89 seconds

All tests
    - 47.75 seconds
```

From a memory perspective, this code:

```
import napalm_yang
c = napalm_yang.base.Root()
c.add_model(napalm_yang.models.openconfig_interfaces)
c.interfaces.interface.add('et1')
c.interfaces.interface.add('et2')
c.interfaces.interface.add('et3')
c.interfaces.interface.add('et4')
c.interfaces.interface.add('et5')
c.interfaces.interface.add('et6')
c.interfaces.interface.add('et7')
c.interfaces.interface.add('et8')
c.interfaces.interface.add('et9')
c.interfaces.get(filter=True)
```

Consumes the following amount of RAM:

* regular pyangbind: 135M
* optimized branch: 97M

96M is actually what it takes to just load in memory all the modules (which means the data now requires ~1M vs ~39M)


robshakir:
Thanks for the PR, as discussed. It looks like there's an issue here with the reference path type (support for YANG leafref). I guess this might be because when we set a leafref value in a list key, then its remote value isn't initialised. I'll try and take a look at this in the next few days.

There's also an issue with the field name `_initialized_<name>` since I think you need to call `safe_name` here. I'll try and get some cycles to look here too.

dbarrosop:
This was extremely hard to troubleshoot because there are a lot of try/except catching on errors that are unrelated and masking the real problem :(

dbarrosop:
Managed to make some progress but the problems introduced by this are very esoteric, mostly because XPathhelper is trying to validate references using an XML object instead of the original pyangbind class, which means the dynamic trick doesn't work :(

dbarrosop:
@robshakir do you have any idea how to fix this?

robshakir:
Sorry it took me a bit to look at this.

The issue here seems to be based on the way that `XPathHelper` works. It assumes that valid entities in the tree register themselves such that we build an XML document that has the same structure as the YANG schema, but none of the values. Each element of this has an `obj_ptr` field that points to an entry in the `library` dict which is a keyed set of references to the actual pyangbind class.  There's a fairly complete explanation of this in [this document](http://pynms.io/pyangbind/xpath/).

When we don't initialise containers then this means that we end up with a structure that is not initialised in the tree - such that trying to retrieve a particular element that hasn't yet been initialised is causing a problem. For example, [this test](https://github.com/robshakir/pyangbind/blob/master/tests/xpath/01-list_leaflist/run.py#L195) will fail because in the tree at that point, there are actually no initialised items. That is, the tree looks like this:

```xml
<root obj_ptr="root"/>
```

Now this test is probably easier to fix, because we can just say -- if an element isn't initialised in the datatree then it shouldn't exist - i.e., the test can just initialise that object, which means that the tree changes how it looks and we get the expected test pass:

```python
  def test_get_list_retrieves_correct_attribute(self):
    self.instance.standalone.l.add(42)
    self.assertEqual(self.path_helper.get_list('/standalone/l')._yang_name, 'l')
``` 

And the tree that we see then:

```xml
<root obj_ptr="root">
  <standalone obj_ptr="2940b1fd-678f-11e8-b47f-9801a7ad8cb7">
    <l obj_ptr="2943c2d7-678f-11e8-a0fa-9801a7ad8cb7" x="42">
      <x obj_ptr="2943f287-678f-11e8-9066-9801a7ad8cb7"/>
    </l>
  </standalone>
</root>
```

Looking at it, this is basically what makes all of the tests fail - since the current expected behaviour is that we can find elements that don't exist in the tree.

The reason for this design was that if you build an API around the classes, then it's useful to be able to do (in pseudo-code):

```
obj = get object at /system/config
deserialise JSON from payload into obj
```

However, we could change the API we have to say:

```
obj = get-or-create object at /system/config
deserialise JSON from payload into obj
```

The new `get-or-create` would start at the root, and if it didn't find the entity there, it'd create it. This is relatively easy to add. All users would transition to this new `get-or-create` who needed this new behaviour. Alternatively, we can just make `get` implement `get-or-create`. I rather prefer the former (although it's backwards incompatible) because it maintains your memory savings.

The other test failure in the 01 XPATH tests looks to be a bug in the implementation, `init_code` ends up just trying to specify a referenced path. Let's discuss what you're trying to achieve there, since there are cases where the referenced path is inside a list. I think you can just remove that code, and things should work.

If we agree on this get-or-create type pattern, we can figure out how/where to implement this. Additionally, we can debug where the key references are going wrong. I imagine it's a very similar challenge (where the reference to `../config/keyname` is failing because we don't have the `config` container created to retrieve that leaf).

dbarrosop:
Thanks for the detailed explanation. I do agree that the `get-or-create` pattern is the best way to go forward. This week I am quite busy but feel free to push changes to this branch or just create a new one "stealing" code from here if you prefer. I am not particularly attached to it and I am fine with that ;)

I will rebase though just in case.

dbarrosop:
Unfortunately I am not going to have time this week to look into this so feel free to take over my branch or create a new one if you want to look into it.

PR #160, opened 2018-02-23
Python 3 compatibility

tarkatronic:
This is a culmination of efforts from @dbarrosop, @AdamIsrael, and myself.

* All tests are now passing in Python 2.7, 3.4, 3.5, and 3.6
* Linting has been changed to use flake8, and is passing on all files.
* tox is now being used to bootstrap test environments

I have not yet gone through to address issues raised by @robshakir in #138; that will be my next step.

As I talked about with @robshakir, I would like to get the tests transitioned over to using the `unittest` framework, and I'm considering doing that in the scope of this PR in order to add test coverage calculation, because I have discovered some code which is, without question, untested. This concerns me a bit in that this large change may not yet be finished.


tarkatronic:
Also, I have seen the error @ktbyers mentioned on the generated `bindings` module sometimes not showing up. And it appears that TravisCI has now seen that as well, here: https://travis-ci.org/robshakir/pyangbind/jobs/345357705

Has anybody else seen this? Any thoughts on what might cause this?

One thought I had on it is, I wonder if it's possible to somehow output the bindings module into a `StringIO` object instead of a file, and use that as our import module. I have a feeling this might be some sort of file system write timing issue, but I can't say for sure just yet.

dbarrosop:
I wouldn’t refactor tests in this PR, it’s going to be difficult to understand if changes are due to py3 compatibility, due to the refactoring or due to something else. I’d refactor in a different PR and then rebase this one or just do it afterwards.

tarkatronic:
Yeah that's kinda the direction I was leaning as well. I was considering also doing a similar thing with the linting. As much as it has helped, this PR is pretty freaking large at the moment. It might be best to have a few smaller, more consumable, PRs first for some general cleanup.

tarkatronic:
This should now be fixed, via #192.

PR #161, opened 2018-02-26
Switch the linter to flake8 and fix all linting errors

tarkatronic:
This pulls over a lot of the work from #160, in order to trim down that PR


pyangbind/helpers/identity.py
```diff
@@ -118,11 +120,12 @@ def build_store_from_definitions(self, ctx, defnd):
           # and if not, then push this identity back onto the stack
           unresolved_identities.append(this_id)
           unresolved_identity_count[this_id] += 1
-          if unresolved_identity_count[this_id] > 1000:
-            if unresolved_idc[ident] > 1000:
-              sys.stderr.write("could not find a match for %s base: %s\n" %
-                (iddef.arg, base_name))
-              error_ids.append(ident)            
+          # FIXME(Joey Wilhelm): `unresolved_idc` and `ident` are both undefined. What is the intent of this code?
```
robshakir: I made a comment in the other PR on this -- linking here for completeness: https://github.com/robshakir/pyangbind/pull/160#discussion_r171901037

pyangbind/helpers/misc.py
```diff
@@ -1,5 +1,5 @@
 """
-Copyright 2015, Rob Shakir 
+Copyright 2015, Rob Shakir
```
robshakir: Per the other PR, I'll go through and fix these because this is now (c) all contributors.

pyangbind/lib/yangtypes.py
```diff
@@ -153,7 +153,7 @@ class RestrictedClass(base_type):
       input value is validated against before being applied. The function is
       a static method which is assigned to _restricted_test.
     """
-    #__slots__ = ('_restricted_class_base')
+    # __slots__ = ('_restricted_class_base')
```
robshakir: Looks like this is now just not required, delete the line vs. fix the spacing?

setup.cfg
```diff
@@ -0,0 +1,4 @@
+[flake8]
+ignore=E111,E114,E121,E126,E127,E128,E131
```
robshakir: Could we comment what these different codes are, just so that we're tracking what the style is?

tests/include-import/run.py
```diff
@@ -39,11 +39,12 @@ def main():
   cmd += " %s/%s.yang" % (this_dir, TESTNAME)
   os.system(cmd)
 
-  from bindings import include_import
+  from bindings import include_import  # NOQA: F401
```
robshakir: Could you comment above here as to what the reason for this `NOQA` is please?

tarkatronic: This one is for `module imported but unused`. I left the import there though so that it could be verified as importable. Did you want me to leave an inline comment for this?

As a side note, this will go away completely with the unittest refactor since the bindings generation is done 100% in memory.

pyangbind/helpers/identity.py
tarkatronic: Would it be alright to leave this fix for the py3 branch, so that this branch can stay as just linting?

robshakir: Sure - that works for me.

setup.cfg
robshakir: Thanks!

tests/include-import/run.py
robshakir: This sounds fine with me.

pyangbind/helpers/misc.py
robshakir: No action required from author.

PR #163, opened 2018-03-08
WIP: Convert tests to use unittest

tarkatronic:
This is still a WIP, but given how large this is growing, I thought it might be prudent to open a PR early in order to share and discuss progress.

This will also help maintain a checklist of what is done and what needs to be done still.

Converted tests:

* [X] binary
* [X] boolean-empty
* [X] choice
* [ ] config-false
* [ ] decimal64
* [ ] enumeration
* [ ] extensions
* [ ] extmethods
* [ ] identityref
* [ ] include-import
* [ ] int
* [ ] integration
  * [ ] openconfig-bgp
  * [ ] openconfig-interfaces
* [ ] leaf-list
* [ ] list
* [ ] misc
* [ ] nested-containers
* [ ] notification
* [ ] openconfig-bgp-juniper
* [ ] presence
* [ ] pyvirtualenv
* [ ] rpc
* [ ] serialise
  * [ ] ietf-json-deserialise
  * [ ] ietf-json-serialise
  * [ ] json-deserialise
  * [ ] json-serialise
  * [ ] juniper-json-examples
  * [ ] openconfig-serialise
* [ ] split-classes
* [ ] string
* [ ] submodules
* [X] typedef
* [ ] uint
* [ ] union
* [ ] xpath
  * [ ] 01-list_leaflist
  * [X] 02-static_ptr
  * [X] 03-current
  * [X] 04-root


tests/base.py
```diff
@@ -0,0 +1,38 @@
...
+      raise ValueError("cls.yang_files must be set")
+    pyang_path = distutils.spawn.find_executable('pyang')
+    if not pyang_path:
+      raise RuntimeError("Could not locate `pyang` executable.")
+    base_dir = os.path.dirname(os.path.dirname(__file__))
+    test_path = os.path.dirname(inspect.getfile(cls))
+    yang_files = [os.path.join(test_path, filename) for filename in cls.yang_files]
+    plugin_dir = os.path.join(base_dir, 'pyangbind', 'plugin')
+
+    pyang_cmd = "{} --plugindir {} -f pybind -p {} --use-extmethods --use-xpathhelper {}".format(
```
robshakir: We probably need a bool option here for the `split-class-dir` flag. These base tests just produce a single `bindings.py` but the others generate whole class hierarchies.

tests/base.py
```diff
@@ -0,0 +1,38 @@
...
+      pyang_path, plugin_dir, test_path, ' '.join(yang_files)
+    )
+    bindings_code = subprocess.check_output(
+      pyang_cmd, shell=True, stderr=subprocess.STDOUT, env={'PYTHONPATH': base_dir}
+    )
+    cls.bindings = types.ModuleType('bindings')
```
robshakir: Per the `split-class-dir` point above, we sometimes might need to change the name of the module that's executed here. For example, when we have `--split-class-dir FOO` then the import is `FOO` vs. `bindings`. See [here](https://github.com/robshakir/pyangbind/blob/master/tests/integration/openconfig-bgp/run.py#L117) for an example.

tests/binary/run.py
```diff
@@ -1,131 +1,107 @@
...
+        allowed = True
+        try:
+          self.binary_obj.container.b1 = value[0]
+        except ValueError:
+          allowed = False
+        self.assertEqual(allowed, value[1],
+          "Could incorrectly set b1 to %s" % value[0])
+
+  def test_binary_leaf_default_value(self):
+    default = self.bindings.bitarray("0100")
```
robshakir: What's the logic to call `bitarray` from the `self.bindings` class, rather than simply import it here? (`from bitarray import bitarray`?)

tests/choice/run.py
```diff
@@ -1,92 +1,82 @@
...
+  def test_class_has_container(self):
+    self.assertTrue(hasattr(self.choice_obj, 'container'), "Object does not have container")
+
+  def test_class_has_choice_containers(self):
+    for container in ["case_one_container", "case_two_container"]:
+      with self.subTest(container=container):
+        self.assertTrue(hasattr(self.choice_obj.container, container),
+          "Object does not have choice container %s" % container)
+
+  def test_class_does_not_have_unknown_choices(self):
```
robshakir: Suggest to rename to `test_class_does_not_have_choices_as_attributes` - the intention here is to check that we don't generate `choice-one` which doesn't include in the YANG schema path.

i.e.,

```
container ca {
  choice a-choice {
     case a { 
       leaf ch-one { type string; }
     }
     case b { 
       leaf ch-two { type string; }
    }
  }
}
```

results in schema paths that are `/ca/ch-one` and `/ca/ch-two`, so we want to make sure that the `a-choice`, `a` and `b` statements don't get erroneously translated into an attribute of `ca`.

tests/binary/run.py
tarkatronic: Absolutely none! :D Total oversight on my part. I'll fix that up.

tests/__init__.py
```diff
@@ -0,0 +1,10 @@
+try:
```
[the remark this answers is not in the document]
robshakir: We agreed to leave this for a subsequent PR.

tests/base.py
robshakir: We agreed that this could be covered in a subsequent PR on Slack.

tests/choice/run.py
robshakir: Could you address this one before we merge? I think the existing name is not quite clear.

robshakir:
Thanks for all the changes, merging this PR now.

We can move the list of tests to transition to an issue -- see #164.

PR #166, opened 2018-03-29
Convert the int tests to unittest

tarkatronic:
Given the size of this test file, I'm opening this as an independent PR

### DIFF
--- a/tests/int/run.py
+++ b/tests/int/run.py
@@ -1,317 +1,282 @@
 #!/usr/bin/env python
 
-import os
-import sys
-import getopt
+import unittest
 
-TESTNAME = "int"
+from tests.base import PyangBindTestCase
 
 
-# generate bindings in this folder
-def main():
-  try:
-    opts, args = getopt.getopt(sys.argv[1:], "k", ["keepfiles"])
-  except getopt.GetoptError as e:
-    print(str(e))
-    sys.exit(127)
+class IntTests(PyangBindTestCase):
+  yang_files = ['int.yang']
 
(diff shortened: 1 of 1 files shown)


tests/int/run.py
```diff
@@ -1,317 +1,282 @@
...
+      with self.subTest(leaf=leaf):
+        setter = getattr(self.int_obj.int_container, "_set_%s" % leaf)
+        allowed = True
+        try:
+          setter(value)
+        except ValueError:
+          allowed = False
+        self.assertTrue(allowed, "Could not set int size %s to %d" % (leaf, value))
+
+  def test_set_int_sizes_below_lower_bounds(self):
```
robshakir: Looks like I wrote this the same way as this in the original test, but now you redeclare `bounds`, I recommend moving the `value - 1` from line 209 to just specifying it within the `bounds` dict. This is done in the next test down and it seems cleaner.

tests/int/run.py
```diff
@@ -1,317 +1,282 @@
...
+        allowed = True
+        try:
+          setter(value)
+        except ValueError:
+          allowed = False
+        self.assertTrue(allowed, "Could not set int size %s to %d" % (leaf, value))
+
+  def test_set_int_sizes_above_upper_bounds(self):
+    bounds = {
+      'eight': 2**7 - 1,
```
robshakir: You can just remove the -1 here, since it's being re-added at line 243.

robshakir:
Thanks for making the changes. Merged :shipit:

PR #169, opened 2018-04-02
Testing with remote YANG files

tarkatronic:
I've converted two more sections of tests to unittest:
* ietf-json-deserialize
* juniper-json-examples

For the latter, I had to add the ability to download YANG files from a remote server. I mostly copied over the downloading logic from the old tests, so it maintained all of the retrying and status checking features. In theory, this should work for all of the other integration tests! 🤞

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -2,9 +2,11 @@ import distutils
 import importlib
 import inspect
 import os.path
+import requests
 import shutil
 import subprocess
 import sys
+import time
 import types
 try:
   import unittest2 as unittest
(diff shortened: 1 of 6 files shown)


tests/base.py
```diff
@@ -17,28 +19,58 @@ class PyangBindTestCase(unittest.TestCase):
   pyang_flags = []
   split_class_dir = False
   module_name = 'bindings'
+  remote_yang_files = None
+
+  @classmethod
```
robshakir: (no action required) Thanks for implementing this -- in the future one proposal we had here was either vendoring these files, or importing a known version, but I can implement this later. The current approach means that we get test failures when the source YANG files change, rather than when the library is broken.

tests/base.py
```diff
@@ -17,28 +19,58 @@ class PyangBindTestCase(unittest.TestCase):
...
+  def _fetch_remote_yang_files(cls):
+    # [ {'local_path': str, 'remote_prefix': str, 'files': [str, ...]}, ...]
```
robshakir: Should this comment explain that it describes the format of `cls.remote_files`?

PR #171, opened 2018-04-06
Converted the last of the serialise tests to unittest

tarkatronic:
I also fixed an issue with loading the auto-generated modules via `importlib`. Apparently, there is a module cache which I wasn't clearing out.


tarkatronic:
In my Py3 compatibility efforts, I was seeing a whole bunch of warnings about unclosed file pointers, so I got those fixed in here as well, since they were all related to these serialisation tests.

PR #172, opened 2018-04-10
Convert integration tests to use unittest

tarkatronic:
In addition:
* Updated the .gitignore to include a whole lot more (Largely from github's template)
* Added more cleanup of `sys.modules` necessary for proper use of importlib.

### DIFF
--- a/tests/integration/openconfig-bgp/run.py
+++ b/tests/integration/openconfig-bgp/run.py
@@ -65,7 +65,7 @@ class OpenconfigBGPTests(PyangBindTestCase):
 
   def test_001_add_bgp_neighbor(self):
     self.instance.bgp.neighbors.neighbor.add("192.0.2.1")
-    self.assertGreater(len(self.instance.bgp.neighbors.neighbor), 0)
+    self.assertEqual(len(self.instance.bgp.neighbors.neighbor), 1)
 
   def test_010_filtered_json_output(self):
     self.instance.bgp.global_.config.as_ = 2856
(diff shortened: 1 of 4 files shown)


tests/integration/openconfig-bgp/run.py
```diff
@@ -1,126 +1,71 @@
...
+  ]
+
+  def setUp(self):
+    self.yang_helper = YANGPathHelper()
+    self.instance = self.ocbind.openconfig_bgp(path_helper=self.yang_helper)
 
   def test_001_add_bgp_neighbor(self):
     self.instance.bgp.neighbors.neighbor.add("192.0.2.1")
-    self.assertNotEqual(len(self.instance.bgp.neighbors.neighbor), 0)
+    self.assertGreater(len(self.instance.bgp.neighbors.neighbor), 0)
```
robshakir: Is `assertGreater` quite right here? This seems like it'd not catch a bug whereby the above line added two entries?

tests/integration/openconfig-interfaces/run.py
```diff
@@ -1,132 +1,67 @@
...
+  ]
+
+  def setUp(self):
+    self.yang_helper = YANGPathHelper()
+    self.instance = self.ocbind.openconfig_interfaces(path_helper=self.yang_helper)
 
   def test_001_populated_intf_type(self):
     i0 = self.instance.interfaces.interface.add("eth0")
-    self.assertNotEqual(len(i0.config.type._restriction_dict), 0)
+    self.assertGreater(len(i0.config.type._restriction_dict), 0)
```
robshakir: Same comment as above :-)

PR #173, opened 2018-04-11
More unittest conversions

tarkatronic:
This one converts the `config-false` and `identityref` tests to use unittest

### DIFF
--- a/tests/identityref/run.py
+++ b/tests/identityref/run.py
@@ -1,166 +1,125 @@
 #!/usr/bin/env python
 
-import os
-import sys
-import getopt
+import unittest
 
-TESTNAME = "identityref"
+from tests.base import PyangBindTestCase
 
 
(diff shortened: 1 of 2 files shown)


tests/identityref/run.py
```diff
@@ -1,166 +1,125 @@
...
+
+  def setUp(self):
+    self.instance = self.bindings.identityref()
+
+  def test_identityref_leafs_get_created(self):
+    for leaf in ['id1', 'idr1']:
+      with self.subTest(leaf=leaf):
+        self.assertTrue(hasattr(self.instance.test_container, leaf))
+
+  def test_cant_assign_string_to_identityref(self):
```
robshakir: Suggest `test_cant_assign_invalid_string_to_identityref` - since all `identityref` values are strings, it's simply that this one is not a value specified in the YANG file.

PR #176, opened 2018-04-13
Converted more tests to unittest, and fixed a silly bug

tarkatronic:
This one converted the following tests to use unittest:

* extmethods
* notification
* presence
* split-classes

I also discovered a super simple bug that had me wanting to smash my head on the desk. 😄

### DIFF
--- a/tests/notification/run.py
+++ b/tests/notification/run.py
@@ -17,7 +17,7 @@ class NotificationTests(PyangBindTestCase):
   # Note that these tests need to import from the filesystem rather than using
   #  the bound module, so that they can directly reference the sub-modules
   def test_set_leaf_inside_notification(self):
-    from .bindings.notification_notification.alert_one import alert_one
+    from bindings.notification_notification.alert_one import alert_one
     instance = alert_one(path_helper=self.path_helper)
     allowed = True
     try:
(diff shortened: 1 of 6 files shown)


tests/notification/run.py
```diff
@@ -1,174 +1,82 @@
...
+  split_class_dir = True
+  pyang_flags = ['--use-xpathhelper', '--build-notifications']
+
+  def setUp(self):
+    self.path_helper = YANGPathHelper()
+
+  # Note that these tests need to import from the filesystem rather than using
+  #  the bound module, so that they can directly reference the sub-modules
+  def test_set_leaf_inside_notification(self):
+    from .bindings.notification_notification.alert_one import alert_one
```
robshakir: Looks like this consistently causes a test failure because the test file isn't a module.

tarkatronic:
The build failure on this one is strange. And a bit perplexing. It's building fine on both my Mac and Linux laptops under Python 2.7. Travis, however, is reporting: `ValueError: Attempted relative import in non-package`.

I can fix these imports... but what's perplexing is why it's having this issue.

tarkatronic:
Oh, poop. I just realized why the failures. I've been testing with `coverage run setup.py test -q`, instead of the `tests/run.sh` script. Makes a bit more sense now. So I'll go ahead and fix those!

PR #181, opened 2018-04-16
Convert the string tests to use unittest

tarkatronic:
I also had to rename this from `string` to `strings`. Since this is
using unittest now, along with the `unittest` test loader, the module
name of `string` was colliding with the standard library module of the
same name.


tests/strings/run.py
```diff
@@ -0,0 +1,130 @@
...
+  def test_set_valid_value_on_restricted_string(self):
+    allowed = True
+    try:
+      self.instance.string_container.restricted_string = "aardvark"
+    except ValueError:
+      allowed = False
+    self.assertTrue(allowed)
+
+  def test_set_invalid_value_on_restricted_string(self):
+    with self.assertRaises(ValueError):
```
robshakir: I like this pattern much better than the one that was in the previous code :-)

PR #190, opened 2018-04-21
Minor fixes to the release process.

robshakir:
PTAL and see whether there are other changes we should make here!

Thanks,
r.


tarkatronic:
The one thought I have which might make this easier, and a little friendlier, is that Travis can be set up to do releases for you. As an example, I've got `django-excel-response` set up to that Travis will automagically deploy to PyPi any time I create a tag of the form `x.y.z`. So I just create my releases directly in GitHub and the rest is done for me. :)

Example code here: https://github.com/tarkatronic/django-excel-response/blob/master/.travis.yml#L45

PR #200, opened 2018-05-15
Fixes #179: Ensure uniqueness in leaf-list

### LINKED ISSUE #179: Leaf-list stores multiple duplicate values
Hello,
I noticed in [RFC 6020 section 7.7 for leaf-list](https://tools.ietf.org/html/rfc6020#section-7.7) that there is a requirement on leaf-list which we do not follow in pyangbind.

`The values in a leaf-list MUST be unique.`

### Example:
Given the following yang model:
```yang
module test {
    namespace "http://foo/bar/test/1.0";
    prefix test;
    leaf-list my-leaf-list {
        type string;
    }
}
```
#### Expected
```python
>>> from test import test
>>> from pyangbind.lib.xpathhelper import YANGPathHelper
>>>
>>> graph = test(ph=YANGPathHelper())
>>>
>>> graph.my_leaf_list = ['foo', 'foo']
>>> graph.get()
{'my-leaf-list': [u'foo']}
```
#### Actual
```python
>>> graph.get()
{'my-leaf-list': [u'foo', u'foo']}
```

I think the leaf-list structure might need to resemble a set instead of a list, unless I misunderstood the rfc.

tarkatronic:
Fixes #179 

I couldn't use the `set()` type directly due to the fact they do not maintain order. So instead, I've added a `unique=True` option, along with a uniqueness check on value assignments to `TypedList`.

### DIFF
--- a/pyangbind/lib/yangtypes.py
+++ b/pyangbind/lib/yangtypes.py
@@ -377,19 +377,25 @@ def TypedListType(*args, **kwargs):
     _list = list()
 
     def __init__(self, *args, **kwargs):
+      self._unique = kwargs.pop('unique', False)
       self._allowed_type = allowed_type
       self._list = list()
       if len(args):
         if isinstance(args[0], list):
           tmp = []
           for i in args[0]:
-            tmp.append(self.check(i))
+            if not self._unique or i not in tmp:
+              tmp.append(self.check(i))
           self._list.extend(tmp)
         else:
           tmp = self.check(args[0])
           self._list.append(tmp)
 
     def check(self, v):
+      # Short circuit uniqueness check
+      if self._unique and v in self._list:
+        raise ValueError("Values in this list must be unique.")
+
       passed = False
       count = 0
       for i in self._allowed_type:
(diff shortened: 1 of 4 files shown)


pyangbind/lib/yangtypes.py
```diff
@@ -377,19 +377,25 @@ class TypedList(collections.MutableSequence):
     _list = list()
 
     def __init__(self, *args, **kwargs):
+      self._unique = kwargs.pop('unique', False)
```
robshakir: Is there ever a scenario that we expect that a list is *not* unique? If not, it might be best to make this flag work the other way around such that we have something that says to skip the uniqueness check?

pyangbind/plugin/pybind.py
```diff
@@ -844,7 +844,7 @@ def get_children(ctx, fd, i_children, module, parent, path=str(),
         # TypedList (see lib.yangtypes) with a particular set of types allowed.
         class_str["name"] = "__%s" % (i["name"])
         class_str["type"] = "YANGDynClass"
-        class_str["arg"] = "base="
+        class_str["arg"] = "unique=True, base="
```
robshakir: Having the `unique` the other way around in `TypedListType` would mean that we could avoid this in the generated code?

pyangbind/lib/yangtypes.py
```diff
@@ -377,7 +377,7 @@ class TypedList(collections.MutableSequence):
     _list = list()
 
     def __init__(self, *args, **kwargs):
-      self._unique = kwargs.pop('unique', False)
+      self._unique = kwargs.pop('unique', True)
```
robshakir: Per the comments in Slack. I'm a little torn on this one.

If we make the default here `True`, then we'll end up saying that if existing consumers just upgrade `pyangbind` module from PyPi, but don't regenerate bindings, then existing code that expects this to be the previous state will end up being broken.

Sorry for the hassle, but I think requiring regenerating bindings such that the feature is turned on as you'd had this originally is safer to allow consumers to choose when to enable this.

PR #203, opened 2018-05-18
run _set even if the user sets the default one

dbarrosop:
Right now it's impossible to know if a user tried to set explicitly a value if the value is the default one. For instance, if you try to set `config.enabled = True` whitin an interface `_change()` will always return `False` even though the user explicitly set it. I know that technically it hasn't changed as that's the default value but that makes it hard to build frameworks when you are trying to control certain parameters ignoring others. With this patch we can:

1. Know if the user explicitly set something like `config.enabled = True` byy checking `_change()`.
2. Know if the user set it to the default one by checking the `_default` parameter


tarkatronic:
Could you add a test to verify this updated behavior? Ideally, one which would have failed before this change.

dbarrosop:
done!

dbarrosop:
Not sure why that test is failing. I'd say it's unrelated to my change :S

robshakir:
The other thing that this change will do is make `get` or serialising to JSON include this field even if the field was solely set to the default. This essentially makes pyangbind have the `with-defaults` type approach in [RFC6243](https://tools.ietf.org/html/rfc6243). I think this is OK, but it does result in a change for existing users - albeit it should be a relatively harmless one.

dbarrosop:
Great! Let me rebase my branch.

dbarrosop:
@robshakir everything is green in case you want to merge. If you want some particular change let me know :)

tarkatronic:
@dbarrosop I've got just one more request on this; could you split your test method into two distinct tests, using `self.assertX` methods instead of just `assert` statements? It looks like it should just be all `self.assertTrue`, so a pretty easy change. :)

dbarrosop:
Ok, this one should be ready. Let me know if there is any other change you may want.

tarkatronic:
Awesome, thank you! I'm good with this. It looks like `flake8` is still complaining, but I should be able to force this, since the `black` PR will fix that regardless. :)

tarkatronic:
@robshakir You good with this? I think your review will need to be dismissed or changed to approved before this can be merged.

robshakir:
I'm good -- but not sure we should force merge here. There's going to be a conflict between this PR and the other one anyway - should we merge there first, and then here, or the othe rway around?

The other option is that @dbarrosop merges from the other branch to here, given that both are approved and we just merge once.

dbarrosop:
If you merge the other one I will rebase this.

dbarrosop:
Done! :D

robshakir:
Merged :shipit:. Thanks for the work on these.

raddessi:
Thank you very much!

PR #204, opened 2018-05-19
autoformat code with black

dbarrosop:
Hello,
I was wondering if you were interested in using some autofmt tool a la golang. Right now the code style of pyangbind is "different" :P so using a tool like `black` could solve the problem to begin with and end any future discussions around style. There are plenty of changes here because I ran black on the codebase, the only changes I did to make this work were in the following files:

* tox.ini
* travis.yaml
* requirements.DEVELOPER.txt

Feel free to close if you don't want this. If you want it I can also edit the styling guide in CONTRIBUTING.md

Regards


robshakir:
Hmm, @tarkatronic - can we disable the "PR Quality Review" thing, since this seems like it's probably not going to be super-useful until we've fixed whatever it reports first?

robshakir:
Sure -- I'm happy to fix this up. I'd also like @tarkatronic's review here.

dbarrosop:
Ok, let me know if you want to move forward and I will update the style guide here: CONTRIBUTING.md

I also rebased to fix the tests.

tarkatronic:
It looks like CI is failing now due to flake8 no longer passing. So it comes down to the question: Should we do away with flake8 if we're going to use an auto-formatter instead?

And yeah the code quality reviewer is less useful at the moment than I was hoping for. If only I could have it check but not report. :-/

tarkatronic:
The one change I would like to propose here is using `--line-length 119`, in keeping with current conventions. This should limit the amount of change proposed here, and it feels like a more sensible line length to me than 88, closer to what is being adopted by a great many projects today.

dbarrosop:
Ok:
- reformatted with `--line-length 119`
- updated contributing guidelines
- removed flake

Let me know if there is anything else you want me to fix :)

PR #205, opened 2018-05-21
Include the `openconfig-if-types.yang` in our downloads

tarkatronic:
This fixes the new issues with Travis/tox builds failing.


dbarrosop:
Not sure if my review is worth anything but +1 :P

dbarrosop:
> "by reviewers with write access"

So, no, it's not :P

PR #208, opened 2018-06-07
Update `reserved_name` list to be a comprehensive list for py2 + py3

unsignedint:
Updates the `reserved_name` list.

WARNING. This has the potential to break a lot of existing code if there is code using existing members  with reserved names, e.g. see the tests that were modified using `input` and `type`.

Not sure what to do about it...


unsignedint:
To be fair, I think the reserved names really only matter for top-level (e.g. when you are importing a yang module that happens to be called `input` or `long`) so perhaps we could be more permissive by changing the implementation. However, I thought I'd create this patch anyway as a talking point as I encountered this issue with a yang I was working with.

pyangbind/lib/yangtypes.py
```diff
@@ -30,61 +30,157 @@
...
+    "decimal",
+    "default",
+    "delete",
+    "get",
+    "insert",
+    "local",
+    "pop",
+    "remove",
+    "yang_name",
+}
```
tarkatronic: This is a long enough list, I think it would be worth moving out into a separate file.

tarkatronic: Also, I think we should keep this as a list instead of a set. Since this isn't expected to change in the code, we can ensure uniqueness ourselves.

robshakir:
Thanks for the contribution!

I don't think all the entries in this are equally syntactically problematic. For example:

```python
class test(object):
  as  = 1
```

will always throw an error, `as` is reserved. `import` falls into the same category.

However, a number of other entries `all`, `abs`, `basestring` etc, don't throw the same error, i.e.:

```python
class test(object):
    abs = ""
    all = ""
    any = ""
    apply = ""
    ascii = ""
    bin = ""
    basestring  = ""
```

works just fine. Equally, I can call a module `abs` and not have an issue (yes, there is a name collision, but it is not a interpreter error).

To your second comment, I think that there is an issue if we have a module called `import`, but not an issue if we have one called `long`. Can you provide a bit more colour on what the problem is here before we make a change that is going to affect multiple consumers please?

Thanks!
r.

unsignedint:
An example of `long` causing an issue:

```python
if six.PY3:
  long = int

from . import long

class timeout(PybindBase):
  def __init__(self, *args, **kwargs):
    self.__long = YANGDynClass(base=long.long, is_container='container', yang_name="long", ..., is_config=True)
```

Here the `base=long.long` throws an error because we have `long` as a built-in type.

robshakir:
Ah, I see.

I think the key thing to deal with here would be where it might become a module name -- this means any `container` or `list`, and probably only impacts `split-class-dir` output. The `reserved_name` is generically used everywhere right now, so I think we would want to split that function out and have these used in different contexts.

Thoughts?

JoseIgnacioTamayo:
Hi,

So far, pyangbind has been running the current list of reserved words... I think the discussion about splitting the reserved-words check must be in an issue and not a PR.

JoseIgnacioTamayo:
Closing old issue.

PR #211, opened 2018-06-12
XML serialisation

unsignedint:
I have made a stab at XML serialisation. I think it mostly works according to the IETF spec, though I have not (yet) been able to compare against some sort of reference implementation or test case.

I did not use the XML Path Helper as per the existing XML branch, instead I have built onto the IETF JSON serialiser. The actually encoding changes to support XML are relatively minor, I think, but... I have refactored a large part of the serialisation module to fit them in (without copy+pasting+modifying).

The best way to review is probably to look at the changes in each commit individually.

Open to suggestions for improvement!


pyangbind/lib/serialise.py
```diff
@@ -164,54 +159,161 @@ def default(self, obj, mode="default"):
...
+
+    def yangt_decimal(self, obj):
+        return six.text_type(obj)
+
+    def yangt_empty(self, obj):
+        return [None] if obj else False
+
+
+class pybindJSONEncoder(json.JSONEncoder):
+    serialiser_klass = YangDataSerialiser
```
robshakir: `s/klass/class/`?

pyangbind/lib/serialise.py
```diff
@@ -164,54 +159,161 @@ def default(self, obj, mode="default"):
...
         elif map_val in ["decimal.Decimal"]:
-            return six.text_type(obj) if mode == "ietf" else float(obj)
+            return self.yangt_decimal(obj)
+
+    def yangt_int(self, obj):
+        # for values that are 32-bits and under..
+        return int(obj)
+
+    def yangt_long(self, obj):
+        # don't need to sepcial-case for PY3 because we assign `long = int` in the header
```
robshakir: `s/sepcial/special/`

pyangbind/lib/serialise.py
```diff
@@ -164,54 +160,271 @@ def default(self, obj, mode="default"):
...
+        return long(obj)
+
+    def yangt_identityref(self, obj):
```
robshakir: It's worth adding a comment here that this implementation never gets called with non-IETF serialisation because we have a `string` there and we don't need to modify it. It took me a little to understand that this was the case here.

pyangbind/lib/serialise.py
```diff
@@ -164,54 +160,271 @@ def default(self, obj, mode="default"):
...
+        raise UnmappedItem
+
+    def yangt_decimal(self, obj):
+        # doesn't seem to need a distinction for non-64-bit decimal?
```
robshakir: Not sure I understand this comment.  In the "internal" JSON format that we use here, decimals are just serialised as JSON numbers - and hence float. This is just an implementation choice of that JSON format.

We should either add a comment to this effect or remove this comment.

pyangbind/lib/serialise.py
```diff
@@ -164,54 +160,271 @@ def default(self, obj, mode="default"):
...
+            if emod != obj._defining_module:
+                return "%s:%s" % (obj._enumeration_dict[obj]["@module"], obj)
+        except KeyError:
+            pass
+        return six.text_type(obj)
+
+    def yangt_decimal(self, obj):
+        return six.text_type(obj)
+
+    def yangt_empty(self, obj):
```
robshakir: This isn't quite correct AFAICS. In general, we expect that an `empty` must be true or not set, if the value is False, we should not include the leaf in the serialised output.

pyangbind/lib/serialise.py
```diff
@@ -164,54 +160,271 @@ def default(self, obj, mode="default"):
...
+        return [None] if obj else False
+
+
+class XmlYangDataSerialiser(IETFYangDataSerialiser):
+    """
+    XML can have an empty tag, and a TypedList must be treated specially, here we mark it with a deque
+    """
+
+    def yangt_typed_list(self, obj):
```
robshakir: It would be worth adding some more commentary as to why this we need to return this as a `deque` rather than simply just returning it as a `list` as in the other implementations.

pyangbind/lib/serialise.py
```diff
@@ -164,54 +160,271 @@ def default(self, obj, mode="default"):
...
+
+class EMF(objectify.ElementMaker):
+    """
+    Override ElementMaker to do netconf specific niceties, e.g.
+     - set the namespace
+     - convert underscore to hyphens for tags
+    """
+
+    def __init__(self, namespace=None, nsmap=None):
+        assert namespace or nsmap, "Must set either namespace or nsmap"
```
robshakir: Can we raise a specific error type here rather than `AssertionError` please? This means that it can be caught more cleanly

pyangbind/lib/serialise.py
```diff
@@ -164,54 +160,271 @@ def default(self, obj, mode="default"):
...
+        if namespace:
+            super(EMF, self).__init__(annotate=False, namespace=namespace, nsmap={None: namespace})
+        elif nsmap:
+            super(EMF, self).__init__(annotate=False, namespace=nsmap[None], nsmap=nsmap)
+
+    def __getattribute__(self, tag):
```
robshakir: Can you add some commentary on why this is required? You can use the `yang_name` attribute of any `YANGBaseClass` to get the real name in the module of any entity. This is preferable since there are other name transformations that can be done to the name in `safe_name`.

pyangbind/lib/serialise.py
```diff
@@ -164,54 +160,271 @@ def default(self, obj, mode="default"):
...
+        else:
+            return yname
+
+    @staticmethod
+    def generate_element(obj, parent_namespace=None, flt=False, with_defaults=None):
+        ietf_tree_json_func = make_generate_ietf_tree(pybindIETFJSONEncoder.yname_ns_func)
+        return ietf_tree_json_func(obj, parent_namespace=parent_namespace, flt=flt, with_defaults=with_defaults)
+
+
+class EMF(objectify.ElementMaker):
```
robshakir: Can we perhaps give this a more descriptive name? Pithy names like this are fine if they are only referred to within their immediate context, but I think that this is intended to be wider scoped.

pyangbind/lib/serialise.py
```diff
@@ -164,54 +160,271 @@ def default(self, obj, mode="default"):
...
+                k, nsmap = k
+                E = EMF(nsmap=dict(nsmap))
+                if isinstance(v, deque):
+                    # TypedList (e.g. leaf-list or union-list)
+                    for i in v:
+                        el = E(k, str(i))
+                        parent.append(el)
+                elif isinstance(v, list):
+                    for i in v:
+                        assert isinstance(i, dict)
```
robshakir: Please raise a specific error here.

pyangbind/lib/serialise.py
```diff
@@ -164,54 +160,271 @@ def default(self, obj, mode="default"):
...
+        if namespace:
+            super(EMF, self).__init__(annotate=False, namespace=namespace, nsmap={None: namespace})
+        elif nsmap:
+            super(EMF, self).__init__(annotate=False, namespace=nsmap[None], nsmap=nsmap)
+
+    def __getattribute__(self, tag):
+        return super(EMF, self).__getattr__(tag.replace("_", "-"))
+
+
+class pybindIETFXMLEncoder(object):
```
robshakir: Please add a docstring.

pyangbind/lib/serialise.py
```diff
@@ -164,54 +160,271 @@ def default(self, obj, mode="default"):
...
+
+    @classmethod
+    def generate_xml_tree(cls, module_name, module_namespace, tree):
```
robshakir: This function would benefit from a docstring.

pyangbind/lib/serialise.py
```diff
@@ -164,54 +160,271 @@ def default(self, obj, mode="default"):
...
+    """
+
+    def yangt_typed_list(self, obj):
+        return deque([self.default(i) for i in obj])
+
+    def yangt_empty(self, obj):
+        return None
+
+
+class pybindJSONEncoderBase(json.JSONEncoder):
```
robshakir: Here and below, it'd be useful to clarify for a reader what these different classes are for.

tests/serialise/xml-serialise/ietf-json-serialise.yang
```diff
@@ -0,0 +1,256 @@
+module ietf-json-serialise {
```
robshakir: We should probably name this module per the convention in the other tests -- i.e.,`xml-serialise`.

tests/serialise/xml-serialise/run.py
```diff
@@ -0,0 +1,92 @@
...
+import six
+from bitarray import bitarray
+from lxml import objectify
+
+from pyangbind.lib.serialise import pybindIETFXMLEncoder
+from pyangbind.lib.xpathhelper import YANGPathHelper
+from tests.base import PyangBindTestCase
+
+
+def xml_tree_equivalence(e1, e2):
```
robshakir: Perhaps we can use http://python-xmlunittest.readthedocs.io/en/latest/xmlunittest.html#xml-documents-comparison-assertion for this test -- it seems easier than maintaining an XML equivalence implementation ourselves here?

pyangbind/lib/serialise.py
unsignedint: Hi Rob, to be honest this is quite new for me. You'll need to elaborate a bit more on what the desired behaviour should be. For now, I have basically tried to conform to the existing (previous) implementation which was this:

```python
         elif orig_yangt in ["empty"]:
            if obj:
                return [None] if mode == "ietf" else True
            return False
```

So my reading of this is basically the following truth table:

obj | IETF-JSON | vanilla JSON
--- | --- | ---
truthy | `[None]` | `True`
falsy | `False` | `False`

So I have the following implementation:

IETF-JSON | vanilla JSON
--- | ---
`[None] if obj else False` | `bool(obj)`

_However_, there is a note later in the code where this diverges from the existing implementation....

```python
       elif map_val in ["pyangbind.lib.yangtypes.YANGBool"]:
            if original_yang_type == "empty" and mode == "ietf":
                if obj:
                    return [None]
            else:
                if obj:
                    return True
                else:
                    return False
```

For an IETF empty with `obj` as false then it would fall through this clause. In this proposed PR this _no longer happens_! I *think* this is safe, but I have no idea about the greater context. It appears to work :)

pyangbind/lib/serialise.py
unsignedint: I've just added a note about the use of *deque*. To expand on that: this is really an implementation detail. All I wanted is to be able to identify a typed list in the next phase of encoding. (Because we serialise containers into a list already, I can't use a list again to represent the typed list).

So I was looking for a _different_ type that hasn't been used. One is to use a set which should be more natural (because, e.g. a leaf-list should behave like a set -- unique items) but the set implementation doesn't maintain ordering and I didn't want to write an `OrderedSet` just for this. A `deque`  behaves like a list and maintains ordering so I used it.

Perhaps, we could actually flip this around => make a container a set (since ordering doesn't need to be maintained AFAIK?) and then => typed list as list ?

tarkatronic: This seems a bit forced/contrived. I think this should use a custom declared type instead. For example, ```python
class CustomList(list):
    pass

return CustomList([self.default(i) for i in obj])
```
Of course, I haven't had a chance to look through this entire set of changes so I can't say for sure this is the right option. But using a new data type just to have something different doesn't feel right.

On a tangential note, this puts me in mind of the ORM-style enhancements I was working on recently. I wonder if, first, it wouldn't make sense to do a type revamp, and turn all of the internal types into actual real types with built-in coercion.

unsignedint: Yep, that makes sense. I've replace the use of *deque* now with a list subclass.

The "ORM-style enhancements" sound interesting. Is that on a branch somewhere? But, I think XML serialisation is useful _now_ (at least for me, and probably many others using NETCONF) so I think we could do this first :)

tarkatronic: Cool, I think this is a good step forward.

The ORM stuff is in a very premature state right now, basically just theoretical/experimental. But I think that enhancements to the serialization could potentially help with it, if the types can all get properly abstracted out. And that can help lead to a cleaner library! Anyhow, if you want to get a look at those changes, they are here: https://github.com/robshakir/pyangbind/compare/master...tarkatronic:orm_refactor and if you want to talk further on them, feel free to join us on the NetworkToCode slack!

pyangbind/lib/serialise.py
robshakir: The issue is `empty` as a YANG type doesn't really have a bool representation. It's either present or not. In the generated classes, the choice was to represent it as a bool and only allow it to be set to true. So before, when the serialisation happened then we would fall through and cause a serialisation error.

Right now, the implementation doesn't prevent someone setting an empty to `False` at set-time, so this seemed a reasonable compromise. I think we should keep this as an error in the serialisation code either way.

For background of why `empty` works like this, it's really an XML element that is:

`<leaf />`

Clearly, this thing is either there (value nothing, it's there) or not.

tests/serialise/xml-serialise/run.py
robshakir: OK, sure! Let's go with this one then :-)

pyangbind/lib/serialise.py
robshakir: Thanks for this change, I agree with Joey's assessment here.

PR #214, opened 2018-06-17
xml-serialisation: fix bug when there is an unset/missing identityref

unsignedint:
Fixes serialisation bug when an identityref leaf hasn't been assigned a value. See test case.

### DIFF
--- a/pyangbind/lib/serialise.py
+++ b/pyangbind/lib/serialise.py
@@ -359,8 +359,8 @@ class pybindIETFXMLEncoder(object):
     def yname_ns_func(parent_namespace, element, yname):
         # to keeps things simple, we augment every key with a complete namespace map
         ns_map = [(None, element._namespace)]
-        if element._yang_type == "identityref" and element in element._enumeration_dict:
-            # valid identityref (i.e. points to a valid identity)
+        if element._yang_type == "identityref" and element._changed():
+            # configured identityref (i.e. points to a valid identity)
             ns_map.append(
                 (element._enumeration_dict[element]["@module"], element._enumeration_dict[element]["@namespace"])
             )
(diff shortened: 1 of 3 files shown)


pyangbind/lib/serialise.py
```diff
@@ -359,7 +359,8 @@ def aux(parent, root):
     def yname_ns_func(parent_namespace, element, yname):
         # to keeps things simple, we augment every key with a complete namespace map
         ns_map = [(None, element._namespace)]
-        if element._yang_type == "identityref":
+        if element._yang_type == "identityref" and element in element._enumeration_dict:
```
robshakir: Minor issue here -- we just detect whether something is set based on using whether the element's value is in its restriction dictionary. Do we really want to do this rather than check `_changed` to see whether it's a non-nil value, and then determining whether to serialise that way?

This code would not error in the case that the `identityref` is set to a value that isn't allowed (which should be prevented by the validation of course). It seems like:

```python
if element._yang_type == "identityref" and element._changed():
  if element not in element._enumeration_dict:
    raise ValueError("identityref %s set to invalid value %s" % (yname, element))
  ns_map.append(...)
```

What do you think?

tests/serialise/xml-serialise/ietf-xml-serialise.yang
```diff
@@ -175,6 +175,12 @@ module ietf-xml-serialise {
         }
       }
 
+      leaf missing-identityref {
```
robshakir: `s/missing/unset/` -- "missing" sounds like the identityref doesn't have a valid base.

unsignedint:
Thanks for your comments, thats a nice improvement.

I've also identified and fixed a bug with the XML equivalence check which meant that missing elements from the serialised result would not be picked up >_< Anyway, I want to look at deserialisation soon so hopefully that means we can drop that XML comparison implementation.

PR #216, opened 2018-07-01
XML deserialisation

unsignedint:
Round-trip test working. Based on the IETF JSON deserialisation method, but with simplifications.
There are probably some scenarios that don't work correctly, but a useful starting point..


pyangbind/lib/serialise.py
```diff
@@ -444,6 +444,112 @@ def generate_ietf_tree(obj, parent_namespace=None, flt=False, with_defaults=None
...
+            if path_helper is not None:
+                # check that this path doesn't already exist in the
+                # tree, otherwise we create a duplicate.
+                existing_objs = path_helper.get(tmp._path())
+                if len(existing_objs) == 0:
+                    obj = base_mod_cls(path_helper=path_helper, extmethods=extmethods)
+                elif len(existing_objs) == 1:
+                    obj = existing_objs[0]
+                else:
+                    raise pybindJSONUpdateError("update was attempted to a node that " + "was not unique")
```
robshakir: Maybe use a different exception name -- `pybindJSONUpdateError` might be confusing given that this is XML. Perhaps we just need a common decode error?

pyangbind/lib/serialise.py
```diff
@@ -444,6 +444,112 @@ def generate_ietf_tree(obj, parent_namespace=None, flt=False, with_defaults=None
...
+                if hasattr(chobj, "_presence"):
+                    if chobj._presence:
+                        chobj._set_present()
+
+                pybindIETFXMLDecoder.load_xml(
+                    child, None, None, obj=chobj, path_helper=path_helper, extmethods=extmethods
+                )
+
+            elif chobj._yang_type == "list":
+                # TODO: development assertions, not sure how to resolve these yet
```
robshakir: I'd raise an exception here rather than just assert -- perhaps just a `UnimplementedError`?

pyangbind/lib/serialise.py
```diff
@@ -444,6 +444,112 @@ def generate_ietf_tree(obj, parent_namespace=None, flt=False, with_defaults=None
...
+                else:
+                    nobj = chobj[key_str]
+
+                # now we have created the nested object element, we add other members
+                pybindIETFXMLDecoder.load_xml(
+                    child, None, None, obj=nobj, path_helper=path_helper, extmethods=extmethods
+                )
+
+            elif hasattr(chobj, "_pybind_generated_by") and chobj._pybind_generated_by == "TypedListType":
+                chobj.append(str(child))
```
robshakir: Doesn't this case every type to string within a `TypedListType`? Does it work with `e.g., leaf-list foo { type union { type uint32; type string; }}`? I'd expect that in this case we can have an XML payload that is simply a `uint32`, whereas this looks like everything would be added to the list as a string.

tests/serialise/xml-deserialise/ietf-xml-deserialise.yang
```diff
@@ -0,0 +1,274 @@
...
+            type string;
+          }
+        }
+        case two {
+          leaf two-leaf {
+            type string;
+          }
+        }
+      }
+      leaf-list ll {
```
robshakir: Please add a testcase that covers a non-string leaf-list :-)

pyangbind/lib/serialise.py
unsignedint: I've renamed the exception. Naming is hard. Let me know if you have a better suggestion.

pyangbind/lib/serialise.py
unsignedint: Great point. This turns out to be rather tricky, particularly with XML, as there is no real typing information.. this isn't a problem with the YANG as a schema.. except in the case of `union`s.

I've had an initial stab by relying on the lxml guess on matching Python type, which I think is a _reasonable_ assumption for minimal effort. I've added a fallback to stringify if and only if string is an accepted type. (The main use-case for this is when you have modelled a leaf as a string but are actually storing ints or floats (e.g. vlan ID, or some QoS priority perhaps..?)

I would like to suggest we go with this and have a separate task to improve and robustify following proper spec (e.g. section 9 RFC 7950).

PR #217, opened 2018-07-03
fix: serialisation element ordering for py<3.6

unsignedint:
Found an issue where serialised XML output would have incorrect ordering _sometimes_. It turns out I never saw this because I was developing with python 3.6, which features an ordered `dict` implementation.

I have made an update to explicitly use `OrderedDict` instead, so we have the same behaviour for python <3.6.

### DIFF
--- a/pyangbind/lib/serialise.py
+++ b/pyangbind/lib/serialise.py
@@ -105,7 +105,7 @@ class YangDataSerialiser(object):
             return [self.default(i) for i in obj]
         # Expand dictionaries
         elif isinstance(obj, dict):
-            return {k: self.default(v) for k, v in six.iteritems(obj)}
+            return OrderedDict((k, self.default(v)) for k, v in six.iteritems(obj))
 
         if pybc is not None:
             # Special cases where the wrapper has an underlying class
@@ -407,7 +407,7 @@ def make_generate_ietf_tree(yname_ns_func):
             # were a scalar.
             return obj
 
-        d = {}
+        d = OrderedDict()
         for element_name in obj._pyangbind_elements:
             element = getattr(obj, element_name, None)
             yang_name = getattr(element, "yang_name", None)


unsignedint:
To be honest, I think this patch probably obviates the need for the structural XML document equality test I was doing to provide serialised outputs were the same. It turns out, I think, ordering does matter and we may as well be consistent about it.

pyangbind/lib/serialise.py
```diff
@@ -407,7 +407,7 @@ def generate_ietf_tree(obj, parent_namespace=None, flt=False, with_defaults=None
             # were a scalar.
             return obj
 
-        d = {}
+        d = OrderedDict()
```
robshakir: `obj._pyangbind_elements` doesn't have a strict requirement for ordering today (it will get defined based on the ordering in the schema). However, I think the reason to implement this is to have keys of lists ordered first -- which isn't guaranteed. For example:

```yang
module bar {
	prefix "b";
	namespace "urn:b";

	list foo {
		key "baz bat";
	
		leaf bat { type string; }
		leaf baz { type string; }
	}
}
```

Results in the output:

```
▶ grep "pyangbind_elements" binding.py | grep bat
  _pyangbind_elements = OrderedDict([('bat', bat), ('baz', baz), ])
```

But I think that you want 'baz' to be output first, since RFC6020 says:

```
   The list's key nodes are encoded as subelements to the list's
   identifier element, in the same order as they are defined within the
   "key" statement.
```

So, some pre-processing is likely required here AFAICS.

JoseIgnacioTamayo:
Hi,

Could you please rebase to a recent versions of pyangbind, and python >= 3.7 (lowest automatically-tested version) ?

Thanks.

JoseIgnacioTamayo:
Closing old Issue. Pyangbind has moved to Python >= 3.7

PR #220, opened 2018-08-13
fix a identityref field module name namespace missing issue.

guojunzhang:
the issue is: when load identityref field to Yang data tree with module
name as the namespace, when we dump the Yang data to ietf json, the
namespace will lost.

the root cause is: when we generate python code from Yang module, the
restriction_arg for the field doesn't include module name as the prefix.

the fix is: add source_module into the imported_prefixes, so pyangbind can
generate also generate the module name as the namespace for identityref field.

### DIFF
--- a/pyangbind/helpers/identity.py
+++ b/pyangbind/helpers/identity.py
@@ -31,6 +31,8 @@ class Identity(object):
         self.children = []
 
     def add_prefix(self, prefix):
+        if self.source_module not in self._imported_prefixes:
+            self._imported_prefixes.append(self.source_module)
         if prefix not in self._imported_prefixes:
             self._imported_prefixes.append(prefix)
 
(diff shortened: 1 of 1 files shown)


guojunzhang:
I signed it!

robshakir:
Thanks for the fix here.

Can we add a test case that covers this please?

fperrin:
Hi @xavier-contreras & @JoseIgnacioTamayo ,

I've hit the same issue, so I'd love to see this fix included in pyangbind :)

The following change adds a testcase that triggers the issue described by @guojunzhang : https://github.com/fperrin/pyangbind/commit/8f8a70e5c49acbb9bdcf506d8b1a832cda008d78. I can open a new PR, or you can merge this one and then commit that UT.

JoseIgnacioTamayo:
Hi,

@fperrin , feel free to propose a new PR with the change + tests.

Thanks.

fperrin:
> @fperrin , feel free to propose a new PR with the change + tests.
> 
> Thanks.

Hi @JoseIgnacioTamayo , did just so, as #322 .

JoseIgnacioTamayo:
Fixed in https://github.com/robshakir/pyangbind/pull/322

PR #222, opened 2018-08-23
Fix _is_config attribute for non-config containers

nicolasmorini:
When a YANG definition of a container has the statement "config false;",
the Python bindings generated for this container should have "is_config=False"
in its __init__ method, but it was not the case.

Containers nested in non-config containers correctly inherited the _is_config
attribute from its parent.

Add two test cases to tests/config-false.

### DIFF
--- a/pyangbind/plugin/pybind.py
+++ b/pyangbind/plugin/pybind.py
@@ -1463,12 +1463,14 @@ def get_element(ctx, fd, element, module, parent, path, parent_cfg=True, choice=
                 register_paths=register_paths,
             )
 
+            elemconfig = class_bool_map[element.search_one("config").arg] if element.search_one("config") else True
+
             elemdict = {
                 "name": safe_name(element.arg),
                 "origtype": element.keyword,
                 "class": element.keyword,
                 "path": safe_name(npath),
-                "config": True,
+                "config": elemconfig,
                 "description": elemdescr,
                 "yang_name": element.arg,
                 "choice": choice,
--- a/tests/config-false/run.py
+++ b/tests/config-false/run.py
@@ -33,6 +33,9 @@ class ConfigFalseTests(PyangBindTestCase):
             allowed = False
         self.assertFalse(allowed)
 
+    def test_container_reports_not_configurable_with_config_false(self):
+        self.assertFalse(self.test_instance.container.subtwo._is_config)
+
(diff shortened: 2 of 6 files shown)


nicolasmorini:
I signed it!

nicolasmorini:
I registered my work email address!

pyangbind/plugin/pybind.py
```diff
@@ -1463,12 +1463,14 @@ def get_element(ctx, fd, element, module, parent, path, parent_cfg=True, choice=
                 register_paths=register_paths,
             )
 
+            elemconfig = class_bool_map[element.search_one("config").arg] if element.search_one("config") else True
```
tarkatronic: As a general practice, I try to avoid duplicated function calls like this, especially in a single line, just to avoid the additional computational overhead. And to improve readability a bit. I think this might read a bit better as
```python
elemconfig = element.search_one("config")
elemconfig = class_bool_map[elemconfig.arg] if elemconfig else True
```

tarkatronic:
Aside from the one comment, the code LGTM. I'll wait on @robshakir for approval of the behavior. 👍

pyangbind/plugin/pybind.py
nicolasmorini: Hi @tarkatronic , thanks for the prompt review!

To be honest, I didn't pay a lot of attention to the code itself, I just focused on the desired functionality. That said, I just copied that one-liner from line 1632 of the same file (where YANG leaves are processed). So, now we have the same one-liner in two parts of the same function (get_element)

After your comment I'm thinking I should define a function to implement this functionality instead of duplicating the code. On the other hand, I'm not sure that in this case it would provide any benefits.

What's your take in this?

JoseIgnacioTamayo:
Hi,

Thanks.

nicolasmorini:
Hi @JoseIgnacioTamayo!

Unfortunately I have moved on from YANG/Netconf and I lost most of the context for this pull request.

Feel free to take over it if you find it useful, or to just close it altogether.

Best regards,
Nicolas

fperrin:
I can confirm this issue is still present, and at `$work` we have been running with Nicolas' change for a while (thank you @nicolasmorini !). Opened #323 after rebasing over newest master.

JoseIgnacioTamayo:
Fixed in https://github.com/robshakir/pyangbind/pull/323

PR #228, opened 2018-09-18
Changing three dictionaries to OrderDict in serialise.py

ubajze:
Based on the RFC6020, when YANG element is a list, the keys should be the first elements in the XML document (section 7.8.5.). Some devices require a key as the first element in the XML document.

I replaced three regular dictionaries with the OrderDict. My tests show that this is working correctly...

### DIFF
--- a/pyangbind/lib/serialise.py
+++ b/pyangbind/lib/serialise.py
@@ -72,7 +72,7 @@ class YangDataSerialiser(object):
     """
 
     def preprocess_element(self, d):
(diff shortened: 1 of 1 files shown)


ubajze:
I signed it!

ubajze:
I saw that two out of three dictionaries were already converted to OrderedDict in pull request #217. 

Any plan to merge that?

ubajze:
Committed with wrong email...

PR #229, opened 2018-09-18
Changed three dicts to OrderDicts

ubajze:
The list elements should include keys as the first elements in the XML document based on the RFC6020 (section 7.8.5.)


ubajze:
Some similar fixes already published in pull request #217. Any plan to merge those changes to master?

Issue #227, opened 2018-09-18
pybindIETFXMLEncoder fails to serialize the object into XML


muffizone:
Hi,

I was trying to convert the python object into XML using the new XMLEncoder but it is failing with an attribute error.

```
In [60]: oclr = openconfig_local_routing()
    ...: rt = oclr.local_routes.static_routes.static.add("192.0.2.1/32")
    ...: rt.config.set_tag = 42
    ...:
    ...:

In [61]: from pyangbind.lib.serialise import pybindIETFXMLEncoder
    ...: print(pybindIETFXMLEncoder.serialise(oclr))
    ...:
    ...:
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-61-1e1f0873410f> in <module>()
      1 from pyangbind.lib.serialise import pybindIETFXMLEncoder
----> 2 print(pybindIETFXMLEncoder.serialise(oclr))

/usr/local/lib/python3.4/dist-packages/pyangbind/lib/serialise.py in serialise(cls, obj, filter, pretty_print)
    378     def serialise(cls, obj, filter=True, pretty_print=True):
    379         """return the complete XML document, as pretty-printed string"""
--> 380         doc = cls.encode(obj, filter=filter)
    381         return etree.tostring(doc, pretty_print=pretty_print).decode("utf8")
    382

/usr/local/lib/python3.4/dist-packages/pyangbind/lib/serialise.py in encode(cls, obj, filter)
    373         tree = ietf_tree_xml_func(obj, flt=filter)
    374         preprocessed = XmlYangDataSerialiser().preprocess_element(tree)
--> 375         return cls.generate_xml_tree(obj._yang_name, obj._yang_namespace, preprocessed)
    376
    377     @classmethod

AttributeError: 'openconfig_local_routing' object has no attribute '_yang_namespace'

In [62]:
```

I was checking the available attributes for oclr object and it does not seem to have any attribute by the name of _yang_namespace. I am guessing this attribute is used for creating the xml namespace for the root element.

brownoldmonk:
I see similar issue as well. Any update on this?

Muna24:
I had the same error, 
after installing pyangbind (pip install pyangbind) and I check the installed files on mycomputer I  find out that we do not have the same version as the one existing on github. to be able to use it what I did was delete pyang (pip uninstall pyangbind), download the folder from github, put it in the same project and it worked

igrush:
[...]

PR #241, opened 2019-02-11
#227 pybindIETFXMLEncoder fails to serialize the object into XML

igrush:
- fix for unknown _yang_namespace in robshakir/pyangbind.

also Comment typo fix.


JoseIgnacioTamayo:
As repo was deleted, deleting pull request.

PR #273, opened 2020-10-27
Update reserved keywords to support python 3.7

taljmars:
`async` is part of python syntax and need to be handled carefully as part of class generation.


taljmars:
@googlebot I signed it!

JoseIgnacioTamayo:
Completed in https://github.com/robshakir/pyangbind/pull/285

PR #291, opened 2022-11-08
Migrate from collections to collections.abc.

xavier-contreras:
Collections Abstract Base Classes are deprecated since Python 3.3 and  removed in 3.10.


JoseIgnacioTamayo (approved):
LGTM

PR #293, opened 2023-03-13
fix: change the collections import from `collections` to `six.moves import collections_abc`

idefixcert:
The problem we face is, the collections does no longer include MutableSequence, it is moved to collections_abc.
The six package has a fix for that.
`from six.moves import collections_abc`

This pull request makes the library also usable with python 3.10


xavier-contreras:
Hello, I'll be closing this PR in favor of https://github.com/robshakir/pyangbind/pull/291

We'll merge that as soon as we clear up some housekeeping.  Thanks for noticing and proposing a fix though!

PR #297, opened 2023-07-09
Fix binary type representation

fperrin:
Hi @xavier-contreras ; @JoseIgnacioTamayo 

It's fantastic you've picked up maintenance for pyangbind! Thank you for stepping up. Now that pyangbind has a new lease of life, I'll submit a series of fixes as and when I clean them up for public consumption. If I can submit those changes in a way that make it easier for you to review, please tell me.

This first set included here fixes representation of the `binary` built-in type. It also fixes the unit-tests by pinning the version of YANG models to a known passing version. (In fact, my opinion is that the YANG models we use should be committed to the repo, instead of fetching them over the network on every run. I can make that change if you agree.) The UT pass on python3.9. I removed python2.7 as I'm pretty sure my usage of `super()` won't work there; I didn't actually run tox with python3.4 to 3.7

If there is interest, I can add GitHub actions to run the existing unit-tests on every PR.

### DIFF
--- a/docs/yang.md
+++ b/docs/yang.md
@@ -52,7 +52,7 @@ PyangBind does not currently try and be feature complete against the YANG langua
 
  **Type**            | **Sub-Statement**   | **Supported Type**      | **Unit Tests**
  --------------------|--------------------|--------------------------|---------------
- **binary**          | -                   | [bitarray](https://github.com/ilanschnell/bitarray)           | tests/binary
+ **binary**          | -                   | [bytes](https://docs.python.org/3/library/stdtypes.html?#bytes)           | tests/binary
  -                   | length              | Supported           | tests/binary
  **bits**            | -                   | Not supported           | N/A
  -                   | position            | Not supported           | N/A
--- a/pyangbind/lib/serialise.py
+++ b/pyangbind/lib/serialise.py
@@ -28,6 +28,7 @@ from __future__ import unicode_literals
 import json
 from collections import OrderedDict
 from decimal import Decimal
+import base64
 
 import six
 from enum import IntEnum
@@ -130,7 +131,7 @@ class YangDataSerialiser(object):
         elif orig_yangt in ["string", "enumeration"]:
             return six.text_type(obj)
         elif orig_yangt in ["binary"]:
-            return obj.to01()
+            return six.text_type(base64.b64encode(obj), "ascii")
         elif orig_yangt in ["decimal64"]:
             return self.yangt_decimal(obj)
         elif orig_yangt in ["bool"]:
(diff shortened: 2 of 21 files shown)


JoseIgnacioTamayo:
Hi Frederic

Thanks for your contribution! Those unit tests needed a fix, by chance I also started taking a look last weekend.

I would suggest we keep this PR exclusively to fixing the tests, not changing Python versions. We are in the process of setting up GitHub Actions, which will take care of running the tests in several python versions.

JoseIgnacioTamayo:
After a look at the changes files, I think we need to split it:

 * One PR would propose changes to the dealing with binary YANG (changes both the library and tests). I see we would be base64 encoding the binary.
 * A separate PR would take care of updating the OC models. I can take care of that. I think it is fine to track the latest OC models from the master branch (I cannot give a rational reason, it just feels itchy to let the code attached to some old branch of the OC repo).

Could you please reduce the scope of the PR to proposing changes to the way Binary values are treated?

fperrin:
Hi,

Reduced the size of the MR to have only the fix to binary type representation

fperrin:
Hi,

Rebased after the recent improvements to the UT (good work @JoseIgnacioTamayo !).

tests/base.py
```diff
@@ -1,4 +1,5 @@
 import distutils
+import distutils.spawn
```
xavier-contreras, line 2: Do we need this import?

xavier-contreras (commented):
Thanks! Just a couple of minor comments.

tests/base.py
fperrin: tl:dr; no, I'll remove that change.

Looks like on python3.9 that I have laying around in an older machine:
```
Python 3.9.2 (default, Feb 28 2021, 17:03:44)
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import distutils; distutils; distutils.__version__
<module 'distutils' from '/usr/lib/python3.9/distutils/__init__.py'>
'3.9.2'
>>> distutils.spawn
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'distutils' has no attribute 'spawn'
>>> import distutils.spawn
>>> distutils.spawn; distutils.spawn.find_executable
<module 'distutils.spawn' from '/usr/lib/python3.9/distutils/spawn.py'>
<function find_executable at 0x7fc388fd1ee0>
```

It appears:

- in python3.9, there is a distutils package coming from Python, and in that version, `distutils.__init__py` does not import spawn, hence the need for that import
- if you have installed setuptools, then the above works:

```
Python 3.9.16 (main, Dec  7 2022, 10:16:11)
[Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import distutils; distutils
>>> distutils
<module 'distutils' (/Users/fperrin/src/pyangbind/env/lib/python3.9/site-packages/setuptools/_distutils/__init__.py)>
>>> distutils.spawn.find_executable
<function find_executable at 0x10188bd30>
```

- distutils will disappear in python3.12: https://github.com/python/cpython/pull/99061
- the syntax `install_requires = file: requirements.txt` in setup.cfg requires a somewhat recent version of setuptools (https://github.com/pypa/setuptools/pull/3253), so between this and the previous point, users have to install/upgrade setuptools _anyway_, so I can drop the import you mention

xavier-contreras: Awesome, thanks for digging that.

xavier-contreras (approved):

PR #306, opened 2023-07-20
Skip known-failing tests, and block CI on fail.

JoseIgnacioTamayo:
This allows the CI to detect and stop breaking changes, while the known-failing tests are fixed (issues raised for that).


xavier-contreras (approved):
Thanks.  My understanding is that we now detect breakage (previously disabled so we could get out of that state) yet disable the known broken tests which we are tracking.

JoseIgnacioTamayo:
@xavier-contreras , yes, after merging this PR we will detect and stop at breaking changes. There is some homework to fix and no longer skip the few unittests that are known to fail.

PR #213, opened 2018-06-15
Updates json ietf serialization to not prefix a property value that i…

markuscraig:
…s already prefixed. This can occur when ietf serializing an instance that was deserialized from a json ietf string.  For example...

```
# deserialize from json-ietf
$ aps = pybindJSON.loads_ietf(OC_EXAMPLE_ACCESS_POINTS, openconfig_access_points, "openconfig_access_points")

# serialize to json-ietf
$ pybindJSON.dumps(aps, indent=None, mode='ietf')
```

PR #313, opened 2023-08-14
Don't prefix a property value that is already prefixed

fperrin:
This is a rebase of @markuscraig's PR #213 over the most recent master (original PR has a conflict when rebasing), plus a testcase that exhibits the issue.

Without the fix, testcase `test_json_ietf_serialise_namespace_handling_remote` fails with:

```
>               self.assertEqual(
                    data["identityref:ietfint"]["ref"],
                    "remote-two:remote-id",
                )
E               AssertionError: 'remote-two:remote-two:remote-id' != 'remote-two:remote-id'
E               - remote-two:remote-two:remote-id
E               ? -----------
E               + remote-two:remote-id
```

Reference: https://www.ietf.org/rfc/rfc6020.txt sections 9.10.*

### DIFF
--- a/pyangbind/lib/serialise.py
+++ b/pyangbind/lib/serialise.py
@@ -235,7 +235,9 @@ class IETFYangDataSerialiser(YangDataSerialiser):
         try:
             emod = obj._enumeration_dict[obj]["@module"]
             if emod != obj._defining_module:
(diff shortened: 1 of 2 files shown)


xavier-contreras (approved):
Thank you, LGTM

Issue #314, opened 2023-08-31
Support `bits` type


tehAgitto:
Hello,

is it possible to support `bits` type? I've tried to convert YANG model into pydantic `BaseModel`, but it seems not every type is supported (I've read docs). Do you have it on your roadmap anywhere?

```python
could not find a match for SOME type -> ['bits']
```

fperrin:
Hi @tehAgitto ,

For `$work` I added support for the bits type. The code is there: https://github.com/fperrin/pyangbind/commits/fred/add-bits

It's working well enough for the models I use, but I never got around to creating a PR to get it included. I think I need to review it.

PR #315, opened 2023-09-04
Handle the "bits" type

fperrin:
The YANG "bits" type gets mapped to a `Set[str]` object in Python. The set object is augmented with a dictionary keeping track of the bit positions. This is used when forming the JSON or XML serialisation.

closes https://github.com/robshakir/pyangbind/issues/314


JoseIgnacioTamayo (approved):
Awesome! Thanks.

I did not like the name 'YANGBitsFactory', but I could not think of a better name for a method that creates a class. So it is fine by me.

fperrin:
That can be renamed to `YANGBitsType` returning `YANGbits` to follow the same pattern as other types. I'll do that now

```python
def RestrictedPrecisionDecimalType(*args, **kwargs):
    """
    Function to return a new type that is based on decimal.Decimal with
    an arbitrary restricted precision.
    """
    precision = kwargs.pop("precision", False)

    class RestrictedPrecisionDecimal(Decimal):
        ...

def YANGListType(*args, **kwargs):
    class YANGList(object):
        ...
```

PR #316, opened 2023-09-04
Optimisation: change class_map deepcopy to copy

fperrin:
This is an improvement by a colleague of mine, @jamie01

pyangbind was taking around 18 seconds on my machine to create the classes for the collection of YANG models for a proprietary routing device. After doing some profiling, deepcopy operations were by far the worst offenders and in particular the deepcopy of "class_map" - a large structure containing all the yang types seen along the way.

Since the copied var doesn't modify deeply nested data, swapping it for a copy seemed reasonable. If I diff the python output file from deepcopy vs copy the files are identical. And not only that, the whole process now completes in under 4 seconds.


JoseIgnacioTamayo (approved):
Awesome! As the tests have not broken, this suggests that the change is fine.

PR #323, opened 2023-10-15
Fix _is_config attribute for non-config containers

fperrin:
Rebase of #222 over latest master. Dropped the UT change, and applied a review suggestion to use a temporary var to avoid fetching the same value twice. Original author is @nicolasmorini .

When a YANG definition of a container has the statement "config false;", the Python bindings generated for this container should have "is_config=False" in its __init__ method, but it was not the case.

Containers nested in non-config containers correctly inherited the _is_config attribute from its parent.

Add two test cases to tests/config-false.

### DIFF
--- a/pyangbind/plugin/pybind.py
+++ b/pyangbind/plugin/pybind.py
@@ -1493,12 +1493,15 @@ def get_element(ctx, fd, element, module, parent, path, parent_cfg=True, choice=
                 register_paths=register_paths,
             )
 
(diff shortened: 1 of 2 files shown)


JoseIgnacioTamayo (approved):
LGTM

miguelbf-alb:
Just faced this issue as well, thanks for the fix which would otherwise take me quite a bit of time!
It'd be nice if it could get in

Issue #324, opened 2023-10-20
pyangbind-0.8.4 installation failed


asn-lns-sa-rj:
Cannot install new version 0.8.4.

$ pip install pyangbind --upgrade
...
    Found existing installation: pyangbind 0.8.3.post1
    Uninstalling pyangbind-0.8.3.post1:
      Successfully uninstalled pyangbind-0.8.3.post1
Successfully installed pyangbind-0.8.4

Contrary to expectation, the folder contains only the __init__.py file, not the subfolders (helpers, lib, plugin)

Then rollback: 
$ pip uninstall pyangbind
$ pip install --upgrade pyangbind==0.8.3.post1
... Successfull

JoseIgnacioTamayo:
Oh my... Something went wrong with the package done and uploaded with GitLab Actions.

I just built and uploaded 0.8.4.post1. It looks OK to me, please give it a try.

Apologies for the inconvenience.

asn-lns-sa-rj:
Thank you for your quick support, Jose.
I confirm it works fine now.
Regards

PR #325, opened 2023-10-22
Fix release automation

JoseIgnacioTamayo:
Intended to prevent https://github.com/robshakir/pyangbind/issues/324 from happening on next releases


xavier-contreras (approved):
Thanks Ignacio!

Issue #305, opened 2023-07-20
Integration and Sample UnitTests using BGP are broken


JoseIgnacioTamayo:
It seems the OC BGP model has changed considerably, the integration and sample tests that use it are failing.

The integration and sample tests that use other models are running and passing fine.

This issue is to update the unittests to match the recent BGP model, or to delete these tests as the goal (integration and sample) is achieved using other OC models other than BGP (like openconfig-interfaces).

PR #330, opened 2023-12-05
Remove Integration tests that use /bgp

JoseIgnacioTamayo:
Remove Integration tests that use /bgp

/bgp was removed since https://github.com/openconfig/public/commit/eabe9951115ddfe310b4913fc429a92732645976

The test objective is acomplished already with openconfig-interfaces model at /interfaces

Fixes https://github.com/robshakir/pyangbind/issues/305


xavier-contreras (approved):

robshakir:
You could move these to use `/network-instances/network-instance/protocols/protocol/bgp` -- since it seems advantageous not to lose coverage. If there's duplication though, seems fine to remove them.

JoseIgnacioTamayo:
Hi there,

Thanks for your inputs.

Although it is good to keep coverage, as I interpret these tests, they just
ensure that an OpenConfig Yang model can be bound and used to produce a
valid JSON blob. This objective is already accomplished with the tests that
use openconfig-interfaces.

With that rationale, I think we can drop these tests.

Saludos / Best regards / Cordialement,

Jose Ignacio Tamayo Segarra
LinkedIn <http://br.linkedin.com/in/joseignaciotamayosegarra>

Issue #336, opened 2024-01-16
Presence container support


ianbarrere:
I came upon [this thread from some years ago](https://github.com/robshakir/pyangbind/issues/70) which indicates that presence containers are now supported in pyangbind, but I can't get my binding to recognize them. The model I'm working with is Cisco-IOS-XR-um-interface-cfg.yang from [here](https://github.com/YangModels/yang/tree/main/vendor/cisco/xr/7921). There is a presence container "shutdown" (why they chose to do it that way is beyond me...) which is not showing up as an attribute when I create my binding. Everything else seems to be recognized.

Is there something I need to do to enable presence containers to be created in the binding when I create it?

Relevant versions:

> pyang              2.6.0
pyangbind          0.8.4.post1

Some output:

```
>>> from network_tools.bindings.xr_um_interfaces import interfaces
>>> interfaces = interfaces.interfaces()
>>> my_interface = interfaces.interface.add("TenGigE0/0/0/1")
>>> my_interface.get()
{'interface-name': 'TenGigE0/0/0/1', 'sub-interface-type': {}, 'ipv4': {'addresses': {'address': {'address': '', 'netmask': '', 'route-tag': 0, 'algorithm': 0}, 'secondaries': {'secondary': OrderedDict()}, 'unnumbered': ''}}, 'ipv6': {'addresses': {'ipv6-address': OrderedDict(), 'link-local-address': {'address': '', 'zone': '', 'route-tag': 0}, 'eui64-addresses': {'eui64-address': OrderedDict()}}}, 'dampening': {'decay-half-life': {'value': 0, 'reuse-threshold': {'value': 0, 'suppress-threshold': {'value': 0, 'max-suppress-time': {'value': 0, 'restart-penalty': {'value': 0}}}}}}, 'encapsulation': {'frame-relay': {}}, 'mtu': 0, 'logging': {'events': {}}, 'bandwidth': 0, 'description': '', 'vrf': '', 'address-family': {'ipv4': {'multicast': {'topologies': {'topology': OrderedDict()}}}, 'ipv6': {'multicast': {'topologies': {'topology': OrderedDict()}}}}}
>>> my_interface.shutdown
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'YANGBaseClass' object has no attribute 'shutdown'
>>>
```

Here is the command I used to generate the binding:

```
[...]

PR #337, opened 2024-02-01
Fix presence with SplitClassDir option, and containing Groups

JoseIgnacioTamayo:
Likely solves https://github.com/robshakir/pyangbind/issues/336

 * Correct encoding of a model instance after a presence container was set/unset.
 * Generate correct python code with SplitClassDir option, when the model has Presence containers
 * Render correctly Presence containers using groupings

### DIFF
--- a/pyangbind/lib/yangtypes.py
+++ b/pyangbind/lib/yangtypes.py
@@ -1090,7 +1090,7 @@ def YANGDynClass(*args, **kwargs):
             except TypeError:
                 super(YANGBaseClass, self).__init__()
             except Exception:
-                six.reraise()
+                raise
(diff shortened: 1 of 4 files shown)


JoseIgnacioTamayo:
Because of black formatting, I had to commit several formatting-only changes for the tests to pass OK.

I could do the formatting changes in another PR, and after merged I will rebase this PR.

But we might avoid the hussle by allowing the formatting changes in this PR (not related to fixing the issue with presence)

JoseIgnacioTamayo:
Let's make things right. 

https://github.com/robshakir/pyangbind/pull/338

PR #339, opened 2024-02-11
Fix presence with SplitClassDir option, and containing Groups

JoseIgnacioTamayo:
Likely solves https://github.com/robshakir/pyangbind/issues/336

- Correct encoding of a model instance after a presence container was set/unset.
- Generate correct python code with SplitClassDir option, when the model has Presence containers
- Render correctly Presence containers using groupings


xavier-contreras (approved):

JoseIgnacioTamayo:
Hi @ianbarrere . I would like to have the currently open PR merged, to make a new release. It should not take long.

PR #344, opened 2024-02-28
Remove six

JoseIgnacioTamayo:
Because there is only Python3 now

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -28,7 +28,5 @@ exclude_lines =
     if 0:
     if __name__ == .__main__.:
 
-    # Don't complain about Py3 compatibility shims
-    if six.PY3:
 
 ignore_errors = True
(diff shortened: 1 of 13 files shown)


xavier-contreras (approved):

JoseIgnacioTamayo:
I would like to get some opinions for this change. Python2 is gone, and we can stop using six as the code works on Python3. Of course, this breaks Python2... which might be OK to break?

xavier-contreras:
IMO this is fine.  Developer survey shows less than 7% still using python2 in 2022.  I'd expect this to be lower now, see second link. 

https://lp.jetbrains.com/python-developers-survey-2022/#PythonVersions

https://www.jetbrains.com/lp/devecosystem-2023/python/ (this shows 1%)

PR #347, opened 2024-04-19
fix: readme next-hop typo

marco-minervino:
This should fix what i think it's a typo as `next-hops` is a container of the list `next-hop`


xavier-contreras (approved):
LGTM

PR #360, opened 2024-09-01
355: yang type instance-identifier imp9lemented with YANGInstanceIden…

sanchayanghosh:
…tifier

Puts the YANGInstanceIdentifier type as an implementation of instance-identifier built-in YANG type.

THis function currently gets and sets the YANG instance identifier methods.


sanchayanghosh:
- [ ] Add instance-identifier tests.
- [ ] Subclass pybindtest for checking the type

PR #363, opened 2025-07-09
Update Python version for CI tests

JoseIgnacioTamayo:
* Keep pyangbind requiring at least Py3.7, but start CI tests at Py3.8
 * Extend CI Tests to Py3.13
 * Update Serialized JSON blobs for tests
 * Lock on setuptools because pyang uses `pkg_resources`
 * Update the integration tests, no longer use openconfig-if-ip as it drags too many dependencies

### DIFF
--- a/tests/integration/openconfig-interfaces/run.py
+++ b/tests/integration/openconfig-interfaces/run.py
@@ -9,8 +9,7 @@ from tests.base import PyangBindTestCase
 
 class OpenconfigInterfacesTests(PyangBindTestCase):
     yang_files = [
-        os.path.join("openconfig", "%s.yang" % fname)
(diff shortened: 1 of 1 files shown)


tests/serialise/openconfig-serialise/json/interfaces_ph.False-flt.False-m.default.json
```diff
@@ -13,7 +13,7 @@
                     }
                 }, 
                 "config": {
-                    "enabled": true, 
+                    "enabled": 1,
```
xavier-contreras, line 16: How come we now have different boolean representations, I would expect these to be:
"enabled": true

